PR #83, opened 2015-07-05
Refactor cam control

stianjensen:
This changes the semantics of the connection between a camerapath and a layer slightly.
The old format matched camerapaths in camerapaths.json with corresponding layers in layers.json by their index in the list. This made it difficult to actually keep a camerapath connected to the proper
layer.

A full rework of the system (diffent camerapaths in different files or something similar) might be best, but that's too much work to be completed before solskogen 2015.

The old format meant that a camerapath was linked to an instance of the layer, i.e. two different instances of the layer could (and would) have different camerapaths. With the new system, the camerapath is instead linked to the 'type' of the layer, as `camerapaths.json` is no longer a list, but a dictionary where the layer type is the key for each dictionary.
This should make it easier to manage the relation between layers and camerapaths, and I doubt the possibility of using different camerapaths for different instances of the same layer will be missed.

### DIFF
--- a/nin/dasBoot/CameraController.js
+++ b/nin/dasBoot/CameraController.js
@@ -31,165 +31,52 @@ CameraController.prototype.parseCameraPath = function(camera_paths) {
   }
 
   if (raw_path.position) {
-    this.position = this.parse3Dkeyframes(raw_path.position);
+    this.position = new PathController(raw_path.position, '3D');
   }
   if (raw_path.lookAt) {
-    this.lookAt = this.parse3Dkeyframes(raw_path.lookAt);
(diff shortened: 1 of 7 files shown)


stianjensen:
Oops, spotted a mistake in example-project. will fix

stianjensen:
Ping

stianjensen:
I tweaked the format a bit further, will test that all of it works properly before this should be merged.

cristeahub:
> I tweaked the format a bit further, will test that all of it works properly before this should be merged.

Was this tested?

sigvef:
Nope.

stianjensen:
I did test it quite a bit more after I commented. Shouldn't be more broken than the previous version

PR #87, opened 2015-07-06
Update three.js in dasboot to r71

stianjensen:


stianjensen:
Apparently this broke the camera fly-around mode. Fix incoming

PR #88, opened 2015-07-06
Add three.js projector

stianjensen:
Moved out from core three.js


cristeahub:
What is this and why do we need it moved out of the core three.js?

stianjensen:
Unclear of me:

It was moved out of the three.js core by the three.js maintainers, so now we need to include it explicitly

PR #94, opened 2015-07-07
Update non-core threejs dependencies

stianjensen:
Should have been updated when updating threejs to r71


aleksanb:
Seems to break stuff in the demo.l

stianjensen:
I'll do some more testing then

nin/dasBoot/lib/ConvolutionShader.js
```diff
@@ -83,7 +83,7 @@ THREE.ConvolutionShader = {
 
 		values = new Array( kernelSize );
 		sum = 0.0;
-		for ( i = 0; i < kernelSize; ++i ) {
+		for ( i = 0; i < kernelSize; ++ i ) {
```
cristeahub: :(

stianjensen: I wouldn't have made that change personally. But this is threejs linting rules ^^

cristeahub: Ye ^^

sigvef:
Does this still break things?

stianjensen:
Probably, since I haven't tested it more. I'll have another look. It would make sense that using the dependencies that came bundled with r71 works better than before, not worse :/

nin/dasBoot/lib/RenderPass.js
```diff
@@ -16,7 +16,7 @@ THREE.RenderPass = function ( scene, camera, overrideMaterial, clearColor, clear
 	this.oldClearAlpha = 1;
 
 	this.enabled = true;
-	this.clear = false;
+	this.clear = true;
```
stianjensen: error logs make it seem like this line might be causing it

stianjensen:
Should be a-ok now

sigvef:
The pixel ratio stuff in the effect composer might tank performance on hdpi laptops, but let's merge this anyway. We should update nin with a resolution selector or something down the line, so we can tackle the problem at that time.

PR #106, opened 2015-07-09
Normalize paths and use node modules for find and cat operations

iver56:
`nin run` should now work as expected on Windows. I haven't tested it on Mac and Linux.


stianjensen:
Works on my machine™

PR #108, opened 2015-07-09
Add customizable start screens

sigvef:


stianjensen:
Load custom demo screens from demo root folder instead, perhaps?

PR #109, opened 2015-07-09
Fix scrolling issues in layers panel

stianjensen:
This should make scrolling both horizontally and vertically in the layers panel a lot better.


aleksanb:
Still double scrollbar in height, but sideways scolling seems better.
Position of the yellow bar is still broken tho (at least in firefox)

stianjensen:
I at least found that scrolling of the left column with layer labels is now in sync with the right column

PR #111, opened 2015-07-09
Remove dbl scroll

stianjensen:
I think I recall testing display: block before, but I don't remember why I decided not to push it.

Anyway, it seems to fix issues I see with layers ending up in the wrong spot in the layers panel

**Bad:**
screenshot 2015-07-09 17 59 43

**Good:**
screenshot 2015-07-09 17 59 17


stianjensen:
Also, notice there's only one scroll bar: 
screenshot 2015-07-09 18 00 30

PR #116, opened 2015-07-12
Exclude Thumbs.db in compiled bin

iver56:


sigvef:
Is there a better way to implement this? I'm not sure we want to maintain a list of things to ignore (Thumbs.db, DS_Store, etc). Can we do something else to make sure Thumbs.db and friends don't appear?

iver56:
Use .gitignore in some way, perhaps?

sigvef:
I like that idea much better.

iver56:
Sure, sounds like a good idea. Would you like to do it?

sigvef:
Ok.

PR #117, opened 2015-07-12
Add convenience root Makefile

sigvef:


iver56:
README.md should be updated accordingly

PR #122, opened 2015-07-12
Add support for multiple concurrent Loaders

sigvef:
This is needed for the live reloading part of the loader.


sigvef:
ping, fixed an issue with layer dependencies

PR #128, opened 2015-07-13
Reload layers when shaders are updated

stianjensen:
Layers can declare dependencies on shaders in layers.json.
They will then be automatically reloaded when a shader they depend on
changes.
This basically means you get shadertoy-like behavior when using nin with
a text editor on your computer!


stianjensen:
shaderreload

PR #133, opened 2015-07-15
Add x scale option for layer view

sigvef:


nin/frontend/app/styles/main.css
```diff
@@ -250,3 +255,98 @@ main-window, main-window > div {
...
+  height: 7.6px;
+}
+input[type=range]:focus::-ms-fill-lower {
+  background: #3d464b;
+}
+input[type=range]:focus::-ms-fill-upper {
+  background: #3d464b;
+}
+
+/* end generated */
```
stianjensen: One day, this will get its own file :sunglasses:

stianjensen:
I have some issues where the yellow marker line disappears sometimes. It seems that `$scope.hideMarker` gets set to true.

I guess the calculation in the bottom of bottom.js should take the scale into account somehow

sigvef:
ping

PR #134, opened 2015-07-19
Account for xScale when editing layer timings

stianjensen:
Editing length of layers after #133 did not work as expected

### DIFF
--- a/nin/frontend/app/scripts/controllers/bottom.js
+++ b/nin/frontend/app/scripts/controllers/bottom.js
@@ -28,17 +28,17 @@
       };
 
       $scope.dragResizeLayer = function(event, ui, layer) {
-        if (ui.position.left != layer.startFrame) {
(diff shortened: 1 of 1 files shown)


sigvef:
We should probably round all these values to whole numbers, as nin doesn't work with layers that start/stop on fractional frame numbers.

stianjensen:
ping

nin/frontend/app/scripts/controllers/bottom.js
```diff
@@ -28,17 +28,17 @@
       };
 
       $scope.dragResizeLayer = function(event, ui, layer) {
-        if (ui.position.left != layer.startFrame) {
+        if (ui.position.left != layer.startFrame * $scope.xScale) {
```
sigvef: You probably also need to round to int here as well.

stianjensen:
ping

PR #135, opened 2015-07-20
Remove twyk, rework camera panel UI

sigvef:
Now values are automatically updated in the UI all the time, instead of
having to click buttons to get the values for each frame. Twyk is at the
same time removed.


stianjensen:
I believe there are some keybindings you can remove, too, which were hooked up to the display of those values

sigvef:
ping

PR #138, opened 2015-07-20
Fail gracefully when camera is undefined

stianjensen:


nin/frontend/app/scripts/services/camera.js
```diff
@@ -63,7 +63,7 @@
           ]) + ',';
         },
         getCameraLookat: function() {
-          var camera = layer.instance.camera;
+          if (!camera) return;
```
sigvef: Might as well use layer.instance.camera directly, in case it changes in the layer?

stianjensen: Can do that. The problem was simply that layer sometimes is undefined

stianjensen:
ping

cristeahub:
lol gracefully :+1:

stianjensen:
https://en.wikipedia.org/wiki/Graceful_exit :wink:

cristeahub:
Nice, didn't know that. But since it doesn't really let the user know what happend I guess it's a "graceful" exit?

stianjensen:
I guess it's actually more like failing silently, then

stianjensen:
Or actually not failing at all, just not doing anything because it doesn't make sense ^^

PR #143, opened 2015-07-25
Add generate to menu

aleksanb:
stufz

We can now generate resources from the frontend.
One more step towards full-fledged IDE.

### DIFF
--- a/nin/backend/generate/TemplateShaderLayer.js
+++ b/nin/backend/generate/TemplateShaderLayer.js
@@ -1,0 +1,24 @@
+/**
+ * @constructor
+ */
+function TemplateLayer(layer) {
+  this.config = layer.config;
+  this.shaderPass = new THREE.ShaderPass(SHADERS.TemplateShader);
+}
+
+TemplateLayer.prototype.getEffectComposerPass = function() {
+  return this.shaderPass;
+};
+
+TemplateLayer.prototype.start = function() {
+};
+
+TemplateLayer.prototype.end = function() {
+};
+
+TemplateLayer.prototype.update = function(frame, relativeFrame) {
+    this.shaderPass.uniforms.time.value = relativeFrame;
+};
+
+TemplateLayer.prototype.resize = function() {
+};
--- a/nin/backend/generate/generate.js
+++ b/nin/backend/generate/generate.js
@@ -10,6 +10,10 @@ var generate = function(type, name) {
     process.exit(1);
   }
 
+  if (name == "") {
+    return;
+  }
+
   var camelizedName = utils.camelize(name);
 
   switch (type) {
--- a/nin/backend/generate/generate.js
+++ b/nin/backend/generate/generate.js
@@ -1,6 +1,7 @@
 var fs = require('fs'),
     path = require('path'),
-    utils = require('../utils');
+    utils = require('../utils'),
+    mkdirp = require('mkdirp');
 
 var generate = function(type, name) {
   var projectRoot = utils.findProjectRoot(process.cwd());
@@ -13,8 +14,27 @@ var generate = function(type, name) {
 
   switch (type) {
     case 'simpleLayer':
-      generateLayer(camelizedName, projectRoot);
-      addToLayers(camelizedName, projectRoot);
+      generateLayer(camelizedName, 'TemplateLayer.js',
+          [[/TemplateLayer/g, camelizedName]],
+          projectRoot);
+
+      addToLayers(camelizedName, 'red', projectRoot);
+      break;
+
+    case 'shader':
+      generateShader(camelizedName, projectRoot);
+      break;
+
+    case 'shaderWithLayer':
+      var shaderLayerName = camelizedName + 'Layer';
+
+      generateShader(camelizedName, projectRoot);
+      generateLayer(shaderLayerName, 'TemplateShaderLayer.js',
+          [[/TemplateLayer/g, shaderLayerName],
+           [/TemplateShader/g, camelizedName]],
+          projectRoot);
+
+      addToLayers(camelizedName + 'Layer', 'pink', projectRoot);
       break;
 
     default:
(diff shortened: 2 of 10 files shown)


stianjensen:
How about just calling it layer instead of simpleLayer?

aleksanb:
The next layer to be added will be a Shader layer, therefore simplelayer to not be confused with others.

sigvef:
How about scene layer then, since they are layers that render a threejs scene?

aleksanb:
Updates coming ahoy!

aleksanb:
Refactored some stuff for more generic layer generation.

Now has support for shader layer && shader generation also.

I renamed Simple layer to just Layer, it turned out to work well!

nin/frontend/app/scripts/controllers/main.js
```diff
@@ -85,6 +85,23 @@
           ]
         },
         {
+          name: 'Generate',
+          items: [
+            {name: 'Layer', click: function() {
+              var layerName = window.prompt("Enter a name for the layer:");
```
sigvef: Prompt blocks js, but does not pause audio. This means that if the demo is running while somebody triggers this menu action, a potentially large number of frames need to be updated and maybe rendered for the js to catch up with the music once the prompt disappears. This can hang the browser, if you are unlucky.

aleksanb:
Code now attributed, tDiffuse removed from shader, demo pauses before generating resource

aleksanb:
Allright Ping

nin/backend/generate/templateShader/uniforms.json
```diff
@@ -0,0 +1,3 @@
+{
+    "time": { "type": "f", "value": null }
```
stianjensen: I think we should have tDiffuse in uniforms for the default shader.

I ran into some very weird issues when I forgot to put it in there.

nin/backend/generate/TemplateShaderLayer.js
```diff
@@ -0,0 +1,24 @@
+/**
+ * @constructor
+ */
+function TemplateLayer(config) {
+  this.config = config;
+  this.shaderPass = new THREE.ShaderPass(SHADERS.TemplateShader);
```
stianjensen: I believe there should be a lowercase 't' in templateShader

nin/backend/generate/TemplateShaderLayer.js
```diff
@@ -0,0 +1,24 @@
+/**
+ * @constructor
+ */
+function TemplateLayer(config) {
+  this.config = config;
```
stianjensen: It might be confusing to call this variable config. The variable that is passed in to the layer is an object containing, amongst other things, `startFrame`, `endFrame`, `config`, etc.

nin/backend/generate/generate.js
```diff
@@ -1,21 +1,66 @@
...
+    case 'shaderWithLayer':
+      var shaderLayerName = camelizedName + 'Layer';
+
+      generateShader(camelizedName, projectRoot);
+      generateLayer(shaderLayerName, 'TemplateShaderLayer.js',
+          [[/TemplateLayer/g, shaderLayerName],
+           [/TemplateShader/g, camelizedName]],
+          projectRoot);
+
+      addToLayers(camelizedName + 'Layer', 'pink', projectRoot);
```
stianjensen: Here, 'Layer' is appended to camelizedName, while on line 25, 'Layer' is not appended.
Intentional?

aleksanb: Yes. This is only called if you generate a Shader + Layer, where you enter the shader name, so if you enter the shader name MyShader in the frontend, it will generate a shader with name MyShader, and a shaderLayer with the name MyShaderLayer.

nin/backend/generate/TemplateShaderLayer.js
aleksanb: It's called config in all our layers and passed in as such in the constructor.
I think it's fair then to call it config.

nin/backend/generate/TemplateShaderLayer.js
aleksanb: DOesn't have to. It's replaced with regex before written out

aleksanb: +           [/TemplateShader/g, camelizedName]],
from generate.js

nin/backend/generate/generate.js
```diff
@@ -1,21 +1,66 @@
...
+      templateShaderPath = path.join(__dirname, 'templateShader');
+
+  mkdirp.sync(targetShaderPath);
+  fs.readdirSync(templateShaderPath).forEach(function (fileName) {
+    var from = path.join(templateShaderPath, fileName),
+        to = path.join(targetShaderPath, fileName);
+    fs.createReadStream(from).pipe(fs.createWriteStream(to));
+  });
+
+  process.stdout.write('Generated shader ' + shaderName + '\n');
```
stianjensen: How does this differ from console.log?

nin/backend/generate/TemplateShaderLayer.js
stianjensen: I'm not aware of any layers where that convention has been used

nin/backend/generate/generate.js
aleksanb: this puts it in stdout, while the other (stderr) in stderr.
More correct for console programs i feel.

nin/backend/generate/TemplateShaderLayer.js
aleksanb: NoiseFXlayer which i based myself on took in config. I see others take in layer.

nin/backend/generate/generate.js
stianjensen: If you look in layers.json, we tend to not append Layer do the end of display name of layers only the type.

The displayname is used in the UI of nin, where we want as little clutter as possible.

nin/backend/generate/TemplateShaderLayer.js
stianjensen: Okay :+1:

nin/backend/generate/TemplateShaderLayer.js
aleksanb: Fixed. I had it correct in TemplateLayer!

aleksanb:
plong

aleksanb:
:D

PR #149, opened 2015-07-26
Add link to online wiki in help menu

stianjensen:
We don't really need the keyboard shortcuts overlay anymore, so this replaces that menu item with something a little bit more useful

### DIFF
--- a/nin/frontend/app/scripts/controllers/main.js
+++ b/nin/frontend/app/scripts/controllers/main.js
@@ -87,7 +87,9 @@
         {
           name: 'Help',
           items: [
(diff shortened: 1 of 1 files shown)


nin/frontend/app/scripts/controllers/main.js
```diff
@@ -87,7 +87,9 @@
         {
           name: 'Help',
           items: [
-            {name: 'Shortcuts', shortcut: '?', click: function() {}}
+            {name: 'Online wiki', shortcut: '', click: function() {
```
aleksanb: Just remove the schortcut thingy!

stianjensen:
ping

PR #151, opened 2015-07-26
Beatbean refactor

stianjensen:


sigvef:
I imagine we can replace the yolo-patching here with initBeatBean as well: https://github.com/ninjadev/nin/blob/cb3b1b6785cbb3923df3b530b15d07bb2cf054e6/nin/backend/compress.js#L80

nin/dasBoot/BEATBEAN.js
```diff
@@ -0,0 +1,31 @@
+(function() {
+  var framesPerBeat;
```
cristeahub: `'use strict'`?

PR #152, opened 2015-07-26
Project settings

stianjensen:


nin/dasBoot/BEATBEAN.js
```diff
@@ -5,8 +5,9 @@
     BEAN = 0;
     BEAT = false;
 
-    var subdivision = 6;
-    var beatsPerMinute = 110 * subdivision;
+    var musicInfo = PROJECT.get('music', {});
+    var subdivision = musicInfo.subdivision || 6;
+    var beatsPerMinute = (musicInfo.bpm || 110) * subdivision;
```
iver56: Slightly inaccurate variable name. BPM (Beats Per Minute) is already the number that is set to 110 in this example. The variable holding the number 110 * 6 (in this example) should be called something different.
How about `subBeatsPerMinute` instead?

sigvef: it is more BEATs per minute, since it is the number of beatbean BEATs, e.g. the thing counted by BEAN.

sigvef:
With this strategy, nin.json can never contain anything secret. Is that ok?

stianjensen:
> With this strategy, nin.json can never contain anything secret. Is that ok?

Interesting, I haven't really considered what other things might be stored in this file. Or that a demo project might be secret, for that matter.

Can you come up with something that one could store in nin.json which could be a bad thing to have end up in the compiled version?

sigvef:
> Can you come up with something that one could store in nin.json which could be a bad thing to have end up in the compiled version?

Credentials for some service that we use in a nin build step, for instance.

nin/dasBoot/BEATBEAN.js
aleksanb: Agree with iver here. BPM has a well-established meaning in things like FL!

aleksanb:
That's probably a concern we can ignore for quite a while (secrets in nin.json!)

nin/dasBoot/BEATBEAN.js
sigvef: I was also agreeing with iver, if that wasn't clear.

stianjensen:
ping

Renamed variables in beatbean.
Also, now only keys which have a default value in projectSettings in the backend will be compiled into the demo. This also gives the guarantee that all PROJECT keys will have a value which makes the API for reading the values nicer.

nin/dasBoot/BEATBEAN.js
```diff
@@ -5,9 +5,8 @@
     BEAN = 0;
     BEAT = false;
 
-    var subdivision = 6;
-    var beatsPerMinute = 110 * subdivision;
-    var beatsPerSecond = beatsPerMinute / 60;
+    var BEATsPerMinute = PROJECT.music.bpm * PROJECT.music.subdivision;
+    var BEATsPerSecond = beatsPerMinute / 60;
```
sigvef: s/beatsPerMinute/BEATsPerMinute/

nin/dasBoot/BEATBEAN.js
```diff
@@ -5,9 +5,8 @@
...
     var framesPerSecond = 60;
     framesPerBeat = framesPerSecond / beatsPerSecond;
```
sigvef: s/beatsPerSecond/BEATsPerSecond/

sigvef:
Consider renaming other variables as well, e.g. framesPerBeat to framesPerBEAT

stianjensen:
ping

PR #156, opened 2015-07-28
Use nin.json as base for project root in all commands

### LINKED ISSUE #155: Definition of project root
Most of the features of the nin backend depend on knowing the root directory of the current project.

Currently, two different methods for determining this are in use.

One, which is used by the serve command, the compiler etc. is to use the current working directory of the process as project root _(i.e. the folder where the nin command was executed)_

The other strategy, used at least by the `nin generate` command, is to check the current directory for the presence of a file called nin.json. If it can not be found in the current directory, the parent directory is searched. This process is repeated until the root of the file system is reached. If no nin.json is present, the nin command terminates.

We should decide on one common way of determing the project root for all parts of nin, to maintain a consistent API, with consistent behavior.

aleksanb:
Fixes #155

### DIFF
--- a/nin/backend/nin
+++ b/nin/backend/nin
@@ -11,41 +11,37 @@ var compile  = require('./compile'),
     utils = require('./utils');
 
 program
-  .command('compile [project]')
+  .command('compile')
   .description('Compile the nin project')
   .option('--no-png-compression', 'Do not use the .png.html compression trick')
-  .action(function(project, options){
-    project = path.resolve(process.cwd(), project || '');
+  .action(function(options){
     console.log(options);
     console.log('compiling project');
-    compile.compile(utils.findProjectRootOrExit(project), {
+    compile.compile(utils.findProjectRootOrExit(process.cwd()), {
         pngCompress: options.pngCompression
     });
   });
 
 program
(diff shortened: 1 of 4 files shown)


nin/backend/nin
```diff
@@ -15,7 +16,7 @@ program
   .action(function(project){
     project = path.resolve(process.cwd(), project || '');
     console.log('compiling project');
-    compile.compile(project);
+    compile.compile(utils.findProjectRootOrExit(project));
```
stianjensen: How does this automatic root finding play together with the optional `project` argument?

aleksanb: I'm not quite sure what project was used for, but it looks like you could run a nin project located somewhere else from where you would run the executable.
As relative path would have to point to the root folder of a nin project to work, nothing breaks, as findProjectRoot will find the nin.json located inside the project.

So the requirement is now more lax, you can now type nin run relative/path/to/project/src/whoops instead of just nin run/relative/path/to/project if you would want to do something like that.

sigvef:
needs rebase

sigvef:
s/User/Use/

aleksanb:
Ping

PR #157, opened 2015-07-28
Add layer module to backend

stianjensen:
Manages adding and updating of layers, and ensures only one write to
layers.json occurs at a time.

### DIFF
--- a/nin/backend/generate/generate.js
+++ b/nin/backend/generate/generate.js
@@ -1,6 +1,7 @@
 var fs = require('fs'),
     path = require('path'),
     utils = require('../utils'),
+    layers = require('../layers'),
     mkdirp = require('mkdirp');
 
 var generate = function(type, name) {
@@ -18,11 +19,22 @@ var generate = function(type, name) {
 
   switch (type) {
     case 'layer':
-      generateLayer(camelizedName, 'TemplateLayer.js',
-          [[/TemplateLayer/g, camelizedName]],
+      var layerName = camelizedName + 'Layer';
+      generateLayer(layerName, 'TemplateLayer.js',
+          [[/TemplateLayer/g, layerName]],
           projectRoot);
 
-      addToLayers(camelizedName, 'red', projectRoot);
+      layers.add(projectRoot, {
+        displayName: name,
+        type: layerName,
+        color: 'red'
+      }, function (err) {
+        if (err) {
+          console.error(err);
+        } else {
+          console.log('Added ' + layerName + ' to layers.json');
+        }
+      });
       break;
 
     case 'shader':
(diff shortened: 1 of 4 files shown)


nin/backend/layer.js
```diff
@@ -0,0 +1,95 @@
...
+function write(projectPath, layers, callback) {
+  var data = JSON.stringify(layers, null, 2) + '\n';
+  fs.writeFile(projectPath + '/res/layers.json', data, function(err) {
+    advanceQueue();
+    callback(err);
+  });
+}
+
+function advanceQueue() {
+  active = operationQueue.shift() || null;
```
aleksanb: Shift gives undefined when array is empty!

nin/backend/layer.js
```diff
@@ -0,0 +1,95 @@
...
+
+function read(projectPath, callback) {
+  operationQueue.push([projectPath, callback]);
+  if (active === null) {
+    advanceQueue();
+  }
+}
+
+function write(projectPath, layers, callback) {
+  var data = JSON.stringify(layers, null, 2) + '\n';
```
aleksanb: We should probably use crossplatform line-endings as this might die on iver's windows machine

nin/backend/layer.js
```diff
@@ -0,0 +1,95 @@
...
+    write(projectPath, layers, function (err) {
+      if (callback) {
+        callback(err);
+      }
+    });
+  });
+}
+
+var operationQueue = [];
+var active = null;
```
aleksanb: Perhaps some more descriptive name? ActiveOperation?

nin/backend/layer.js
```diff
@@ -0,0 +1,95 @@
...
+
+function read(projectPath, callback) {
+  operationQueue.push([projectPath, callback]);
+  if (active === null) {
+    advanceQueue();
+  }
+}
+
+function write(projectPath, layers, callback) {
```
aleksanb: Perhaps something like queueWrite instead?

nin/backend/layer.js
```diff
@@ -0,0 +1,95 @@
...
+      advanceQueue();
+      callback(err);
+    } else {
+      var layers = JSON.parse(data);
+      callback(err, layers);
+    }
+  });
+}
+
+function mergeOptions(input, defaults) {
```
aleksanb: Perhaps move to utils if it might be needed more places?

nin/backend/layer.js
stianjensen: Hence, `|| null`

nin/backend/layer.js
stianjensen: :+1:

nin/backend/layer.js
aleksanb: ah it's set further up. I thought it was only used for the check under, and undefined is falsy.
Wait, shouldn't falsy be equally good as null then?

nin/backend/layer.js
iver56: Nah, it's fine

nin/backend/layer.js
stianjensen: I check explicitly for null here: https://github.com/ninjadev/nin/pull/157/files#diff-b35a718792f937cea8aa69c0155b01afR54

stianjensen:
ping

PR #159, opened 2015-07-29
Use ffmpeg for rendering. Improve readability of render module.

iver56:
Works on Windows \o/
Needs to be tested on OS X and Linux.

### DIFF
--- a/nin/backend/package.json
+++ b/nin/backend/package.json
@@ -9,6 +9,7 @@
     "concat-files": "0.1.0",
     "crc": "3.4.0",
     "express": "4.13.4",
+    "fluent-ffmpeg": "2.0.1",
     "glob": "7.0.3",
     "mkdirp": "0.5.1",
     "readdir": "0.0.13",
--- a/nin/backend/render.js
+++ b/nin/backend/render.js
@@ -1,25 +1,44 @@
 var path = require('path');
 var projectSettings = require('./projectSettings');
-var spawn = require('child_process').spawn;
+var ffmpeg = require('fluent-ffmpeg');
 
 
 var render = function(projectPath) {
-  var musicPath = path.join(projectPath, projectSettings.load(projectPath).music.path),
-      avconv = spawn('avconv', [
-  '-y',
-  '-r', '60',
-  '-i', projectPath + '/bin/render/%07d.png',
-  '-i', musicPath,
-  '-c:v', 'libx264',
-  '-c:a', 'copy',
-  '-crf', '16',  
-  projectPath + '/bin/render/render.mp4']);
-  avconv.stdout.on('data', function(data) {
-    console.log(data.toString());
-  });
-  avconv.stderr.on('data', function(data) {
-    console.log(data.toString());
-  });
-}
+  console.time('Total execution time');
+
+  var pngPath = path.join(projectPath, '/bin/render/%07d.png');
+  var musicPath = path.join(projectPath, projectSettings.load(projectPath).music.path);
+  var destination = path.join(projectPath, '/bin/render/render.mp4');
+
+  ffmpeg()
+    .addInput(pngPath)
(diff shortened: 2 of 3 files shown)


nin/backend/render.js
```diff
@@ -1,21 +1,42 @@
...
+  ffmpeg()
+    .addInput(pngPath)
+    .withInputFps(60)
+
+    .addInput(musicPath)
+    .audioCodec('copy')
+
+    .videoCodec('libx264')
+    .withOutputFps(60)
+    .withOutputOption('-pix_fmt', 'yuv420p') // YUV color space
```
stianjensen: So this is the same as the previous `-y`?

iver56: No, `-y` means overwrite if the file already exists. It seems that I don't need to specify this explicitly, because it doesn't complain about any existing file.

`pix_fmt`, however, specifies the pixel format. yuv420p is made with human perception in mind, so that's good.

stianjensen: So we didn't specify pix_fmt before, or was it default, mayhaps?

iver56: Correct, pix_fmt wasn't specified before. I don't know what the default was/is, but yuv420p seems to be best practice. See also: https://trac.ffmpeg.org/ticket/2214

sigvef:
This didn't work for me, I'll do more testing and report back.

iver56:
What was the error?

Btw, how about we keep both pieces of code and create two separate commands for it (for example `nin render` and `nin render --avconv` for `avconv` and `nin render --ffmpeg` for `ffmpeg`)? Then people can choose whatever floats their boat.

stianjensen:
> Btw, how about we keep both pieces of code and create two separate commands for it 

We could do that, but if we get ffmpeg to work for everyone, that's better, because it's less code to maintain.

iver56:
rebased

nin/backend/package.json
```diff
@@ -9,6 +9,7 @@
     "concat-files": "0.1.0",
     "crc": "3.4.0",
     "express": "4.13.4",
+    "fluent-ffmpeg": "^2.0.1",
```
stianjensen: Could you lock the version number to one specific version, so our asses won't get bitten in the future?

sigvef: Consider removing the `^` for safer version locking.

nin/backend/render.js
```diff
@@ -1,25 +1,44 @@
...
-  });
-  avconv.stderr.on('data', function(data) {
-    console.log(data.toString());
-  });
-}
+  console.time('Total execution time');
+
+  var pngPath = path.join(projectPath, '/bin/render/%07d.png');
+  var musicPath = path.join(projectPath, projectSettings.load(projectPath).music.path);
+  var destination = path.join(projectPath, '/bin/render/demo.mp4');
```
sigvef: render.mp4 (or change the name in the readme to demo.mp4).

iver56: done

nin/backend/package.json
iver56: done

sigvef:
Other than the comments this works fine on my system, and the video render is comparable (but not exactly equal to) the old render.

stianjensen:
> the video render is comparable (but not exactly equal to) the old render.

What are the differences?

stianjensen:
Will take a look at this now.

stianjensen:
LGTM

cristeahub:
>> the video render is comparable (but not exactly equal to) the old render.

> What are the differences?

Did this get answered?

stianjensen:
> > > the video render is comparable (but not exactly equal to) the old render.

> > What are the differences?

> Did this get answered?

Some encoder settings were slightly different, but nothing that was visible when looking at the rendered video.

PR #161, opened 2015-07-30
Bottom panel UI redesign

sigvef:
image

---

image

---

image


cristeahub:
When double clicking on a layer that is not the first layer, the offset of the line is wrong.

sigvef:
ping

PR #163, opened 2015-07-30
Add pathVisualization to dasBoot, pathController

aleksanb:
pathvisualiztion

showhide

Nintage now has support to on-demand generate camerapath- and any other path visualizations.
This is pretty cool to check out where the camera actually goes!

### DIFF
--- a/nin/dasBoot/LayerManager.js
+++ b/nin/dasBoot/LayerManager.js
@@ -49,9 +49,13 @@ LayerManager.prototype.loadLayer = function(layer) {
 LayerManager.prototype.update = function(frame) {
   this.updateActiveLayersList(frame);
   for(var i = 0; i < this.activeLayers.length; i++) {
-    var relativeFrame = frame - this.activeLayers[i].startFrame;
-    this.activeLayers[i].instance && this.activeLayers[i].instance.update(
-        frame, relativeFrame);
+    var layer = this.activeLayers[i],
+        li = layer.instance,
+        relativeFrame = frame - layer.startFrame;
+
+    if (li) {
+      li.update(frame, relativeFrame);
+    }
   }
 };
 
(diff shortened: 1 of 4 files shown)


iver56:
How about a single menu item that toggles the visibility of the camera path? Just like we wouldn't have two different buttons for starting and pausing the demo.

aleksanb:
In a perfect world, the button would toggle between "Show path" and "Hide path", but the menu doesn't currently support that, so i found this more pleasing than one button.
I can change it if people think the toggle is better.

nin/dasBoot/LayerManager.js
```diff
@@ -49,9 +49,21 @@ LayerManager.prototype.loadLayer = function(layer) {
...
+        relativeFrame = frame - layer.startFrame;
+
+    if (li) {
+      li.update(frame, relativeFrame);
+
+      if (li.cameraController &&
+          li.cameraController.position &&
+          li.cameraController.position.visualizer &&
+          frame >= layer.startFrame && frame <= layer.endFrame) {
+            console.log("HIT");
```
cristeahub: Do we need this one?

cristeahub:
I like toggle

nin/dasBoot/LayerManager.js
```diff
@@ -124,6 +136,22 @@ LayerManager.prototype.updateActiveLayersList = function(frame, forceUpdate) {
   }
 };
 
+LayerManager.prototype.showCameraPathVisualizations = function(shouldShow) {
+  for(var i = 0; i < this.layers.length; i++) {
```
iver56: The linter complains about this code style.
https://github.com/ninjadev/nin#linting

aleksanb: I can't see in what build step the linter is used?
It doesn't complain during project compilation.

aleksanb:
Toggle it is then.

nin/dasBoot/LayerManager.js
iver56: I don't think that it's a part of the any build step in nin. I have the linter set up in my IDE, so that it monitors my code and I can fix code style and errors on the fly, while I code. I find that much smoother than running a command after I've coded my stuff, only to find all the things that need fixing according to the linter.

nin/dasBoot/LayerManager.js
```diff
@@ -49,9 +49,20 @@ LayerManager.prototype.loadLayer = function(layer) {
...
+        relativeFrame = frame - layer.startFrame;
+
+    if (li) {
+      li.update(frame, relativeFrame);
+
+      if (li.cameraController &&
+          li.cameraController.position &&
+          li.cameraController.position.visualizer &&
+          frame >= layer.startFrame && frame <= layer.endFrame) {
+        li.cameraController.position.visualizer.update(relativeFrame);
```
stianjensen: We probably don't want this code to always run as part of the demo

aleksanb:
I did consider it, but i'm not sure of where the load order for the dasboot files is defined.
Where do i have to include the file after factoring it out?

nin/dasBoot/LayerManager.js
aleksanb: Agree. Perhaps a LayerManager variable set to true if visualizations are to be shown?

This part of the code is also a bit brittle, as skipping in the demo with visualizatinos enabled sometimes breaks the update loop here, apparently update is called before layers are removed or something on skip.

sigvef:
ping \o/

cristeahub:
\o/

PR #164, opened 2015-07-30
Document the --no-png-compression option

iver56:


sigvef:
nin is also self documenting, fwiw (`nin -h`).

PR #172, opened 2015-08-02
Add DJ features to cue points.

aleksanb:
One can now halve and double the loop length with the press of a button!
Remix your prods live infront of a live audience, watch them go wild!

### DIFF
--- a/nin/frontend/app/scripts/controllers/bottom.js
+++ b/nin/frontend/app/scripts/controllers/bottom.js
@@ -72,6 +72,18 @@
         }
       });
 
+      commands.on('multiplyLoopLengthBy', function(amount) {
+        if ($scope.loopEnd === undefined || $scope.loopStart === undefined) {
+          return;
+        }
+
+        var clampedAmount = Math.max(0, amount),
+            loopLength = $scope.loopEnd - $scope.loopStart,
+            newLoopLength = Math.max(1, loopLength * clampedAmount);
+
+        $scope.loopEnd = $scope.loopStart + newLoopLength;
+      });
+
       $scope.$watch('currentFrame', function (nextFrame) {
         if ($scope.loopActive && nextFrame >= $scope.loopEnd) {
           $scope.demo.jumpToFrame($scope.loopStart);
--- a/nin/frontend/app/scripts/controllers/main.js
+++ b/nin/frontend/app/scripts/controllers/main.js
@@ -55,6 +55,12 @@
             {name: 'Set cue point', shortcut: 'g', click: function() {
               commands.setCuePoint();
             }},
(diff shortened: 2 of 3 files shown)


iver56:
How about being able to drag the loop points (and have them snap to BEATs as you drag) instead? That would be more useful, IMO. I need to slightly adjust my selected loop more often than I need to double or halve its length. I wouldn't use nin as a DJ tool.

aleksanb:
It's quite fun to play with halving and doubling the loop length.
You should add dragging of loop points in addition, as both features are pretty cool.

sigvef:
Consider fixing linting errors:

image

aleksanb:
No lint settings defined in the project, we'll take it as we get that merged in?

nin/frontend/app/scripts/controllers/bottom.js
```diff
@@ -72,6 +72,17 @@
...
 
+      commands.on('multiplyLoopLengthBy', function(amount) {
+        if ($scope.loopEnd === undefined || $scope.loopStart === undefined) {
+          return;
+        }
+
+        var clampedAmount = Math.max(0, amount),
+            loopLength = $scope.loopEnd - $scope.loopStart;
+
+        $scope.loopEnd = $scope.loopStart + (loopLength * clampedAmount);
```
sigvef: Perhaps clamp the new loopLength to at least 1, so we wont get a loop that starts and ends on the same frame?

aleksanb:
CLAMPD

nin/frontend/app/scripts/controllers/bottom.js
```diff
@@ -72,6 +72,17 @@
...
 
+      commands.on('multiplyLoopLengthBy', function(amount) {
+        if ($scope.loopEnd === undefined || $scope.loopStart === undefined) {
+          return;
+        }
+
+        var clampedAmount = Math.max(0, amount),
+            loopLength = $scope.loopEnd - $scope.loopStart;
+
+        $scope.loopEnd = Math.max(1, $scope.loopStart + (loopLength * clampedAmount));
```
sigvef: This clamp may not do what you think it does! As it is written, loops cannot end on frame 0. You may have intended to stop loops from starting and ending of the same frame, i.e. a loop length of 0. In that case, you should use something like loopStart + max(1, loopLength * clampedAmount) instead.

aleksanb:
PING

cristeahub:
LGTM, but I think the musicians should double check this before merging.

lionleaf:
Set an upper limit to the loop length   ( = length of music )

Otherwise works well with newest master.

lionleaf:
Also add hotkeys to README!

PR #173, opened 2015-08-02
Make demo music path user configurable through nin.json

aleksanb:
Previously, you had to have a song called music.mp3 in res, now you can
choose yourself, allowing for easy switching if needed.

### DIFF
--- a/nin/backend/projectSettings.js
+++ b/nin/backend/projectSettings.js
@@ -2,6 +2,7 @@ var fs = require('fs');
 
 var defaultSettings = {
   music: {
+    path: "res/music.mp3",
     bpm: 110,
     subdivision: 6
   }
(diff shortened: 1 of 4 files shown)


sigvef:
Convention vs configuration? At least keep res/music.mp3 as a default setting if music.path is unset.

stianjensen:
It is kept as default there

stianjensen:
But you'll need to change it more places, for instance in waveform.js in the frontend

sigvef:
> It is kept as default there

Not if music.path is unset.

aleksanb:
Updated to correct path in frontend too.

sigvef:
backend/render.js also needs an update.

aleksanb:
Okey, so it's kind of difficult to access project settings in backend right now (@stianjensen you want do do a feature exstension?) but one thing this still does is that it moves all those pesky 'res/music.mp3' to a common file!

stianjensen:
I'll update projectSettings in backend, to make it available.

stianjensen:
With #178 you can do
```
var musicPath = projectSettings.load(projectPath).music.path;
```

aleksanb:
PING

PR #202, opened 2016-04-29
Initialize nin project with layer

aleksanb:


sigvef:
Needs updated commit message now that the generated layer is no longer a shader layer.

aleksanb:
Pong.

PR #203, opened 2016-04-29
Fix grunt minor version

aleksanb:
0.4.1 had an obscure bug when building the frontend of the project


stianjensen:
I guess we'd like to lock down the version numbers of all these packages.

aleksanb:
Default install works with the others atm.
It would be a lot of work to fix all versions. Make a separate issue for that?

sigvef:
> Make a separate issue for that?

+1

PR #206, opened 2016-05-01
Check for presence of layer's render method

stianjensen:
LayerManager calls each layer's render method from its render method,
but previously failed if a layer didn't implement it.


aleksanb:
Shouldn't all layers have a render method?

sigvef:
Layers have to expose an EffectComposer Pass (e.g. ShaderPass, RenderPass, etc) that will be used by the LayerManager's EffectComposer to render a given layer. Additionally, Layers *may* perform additional operations at render-time in the optional render method, which gets called before the Pass is executed by the EffectComposer.

PR #208, opened 2016-05-12
Add animationClip support

sigvef:
It is gloriously underspecified at the moment, but it is at least a
starting point!

### DIFF
--- a/nin/dasBoot/LayerManager.js
+++ b/nin/dasBoot/LayerManager.js
@@ -54,6 +54,26 @@ LayerManager.prototype.update = function(frame) {
         relativeFrame = frame - layer.startFrame;
 
     if (li) {
+      if(layer.animationClips) {
+        for(var key in layer.animationClips) {
+          var animationClip = layer.animationClips[key];
+          for(var j = 1; j < animationClip.length; j++) {
+            var keyframe = animationClip[j];
+            if(keyframe.relativeFrame >= relativeFrame) {
+              var from = animationClip[j - 1];
+              var to = keyframe;
+              break;
+            }
+          }
+          if(from) {
+            li[key] = window[to.type](from.value, to.value,
+                                      (relativeFrame - from.relativeFrame) /
+                                      (to.relativeFrame - from.relativeFrame));
+          } else if(to) {
+            li[key] = to.value;
+          }
+        }
+      }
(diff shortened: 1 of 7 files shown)


sigvef:
To use, add something like this to a layer in res/layers.json:

```
"animationClips": {
        "glowAmount": [
            {"relativeFrame": 0, "value": 0, "type": "hold"},
            {"relativeFrame": 80, "value": 0, "type": "hold"},
            {"relativeFrame": 100, "value": 1, "type": "lerp"},
            {"relativeFrame": 150, "value": 0, "type": "smoothstep"},
            {"relativeFrame": 4000, "value": 0, "type": "hold"}
        ],  
        "bikeElevation": [
            {"relativeFrame": 0, "value": 0.5, "type": "hold"},
            {"relativeFrame": 100, "value": 0.5, "type": "hold"},
            {"relativeFrame": 120, "value": 1, "type": "smoothstep"},
            {"relativeFrame": 140, "value": 0, "type": "smoothstep"},
            {"relativeFrame": 160, "value": 1, "type": "smoothstep"},
            {"relativeFrame": 180, "value": 0, "type": "smoothstep"},
            {"relativeFrame": 200, "value": 0.5, "type": "smoothstep"},
            {"relativeFrame": 4000, "value": 0, "type": "hold"}
        ]   
    }
```

sigvef:
This is what it looks like:

image

nin/dasBoot/LayerManager.js
```diff
@@ -54,6 +54,28 @@ LayerManager.prototype.update = function(frame) {
...
+            }
+          }
+          if(from) {
+            li[key] = window[to.type](from.value, to.value,
+                                      (relativeFrame - from.relativeFrame) /
+                                      (to.relativeFrame - from.relativeFrame));
+          } else if(to) {
+            li[key] = to.value;
+          } else {
+            li[key] = from.value;
```
stianjensen: As I read this, it seems that the only case when we reach this else is when both `from` and `to` are undefined. In that case, `from.value` won't work here.

sigvef:
ping

aleksanb:
How about Automation Clips?

sigvef:
Continue discussion in #209.

PR #209, opened 2016-05-14
Add automationClip support

sigvef:
See #208 for previous discussion.

### DIFF
--- a/nin/backend/layers.js
+++ b/nin/backend/layers.js
@@ -36,7 +36,17 @@ function update(projectPath, layerId, layer, callback) {
       return;
     }
     for (var key in layer) {
-      layers[layerId][key] = layer[key];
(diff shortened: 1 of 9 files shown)


sigvef:
ping, now has an automation clip editor UI as well.

sigvef:
image

sigvef:
ping

sigvef:
ping, rebased

aleksanb:
sadface

An undefined clipName might be the case of one? Not sure where $scope.clipName is set.
Can the error be due to using normal html height and width attributes, and not something like ng-height, ng-width?

clippu

sigvef:
I'm closing this since it isn't all to compatible with node-based nin.

sigvef:
s/to/too/

PR #212, opened 2016-05-15
Add error display in nin frontend

sigvef:
image

### DIFF
--- a/nin/dasBoot/loop.js
+++ b/nin/dasBoot/loop.js
@@ -18,10 +18,10 @@ function createLoop(options) {
       that.deltaTime += that.time - that.oldTime;
       that.oldTime = that.time;
       while (that.deltaTime >= frameLength) {
+        that.deltaTime -= frameLength;
         demo.music._calculateFFT();
         updateBeatBean(that.currentFrame);
         update(that.currentFrame++);
-        that.deltaTime -= frameLength;
       }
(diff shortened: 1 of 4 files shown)


stianjensen:
Ohyeah, the good stuff!

sigvef:
image

Looking through old discarded local branches pays off some times :P

stianjensen:
Yeah, I have some of those, I should look through them before this year's demo season. "Old Code Upload Sunday" might not be as catchy as "Technical Debt Tuesday", but maybe we should make it a thing.

nin/dasBoot/loop.js
```diff
@@ -18,10 +18,10 @@ function createLoop(options) {
       that.deltaTime += that.time - that.oldTime;
       that.oldTime = that.time;
       while (that.deltaTime >= frameLength) {
+        that.deltaTime -= frameLength;
```
stianjensen: How does this relate to the error display?

sigvef: If a layer's update method crashes before deltaTime gets adjusted, the demo will try to "catch up on time" by calling update tons of times in a row after the crash gets fixed. This will crash the execution environment if you are unlucky. If deltaTime gets adjusted beforehand, on the other hand, the loop will not think the the demo is behind on schedule and does not attempt to "catch up".

PR #213, opened 2016-05-15
Add barebones ninjascript transpiler

sigvef:
Now we can edit src/MyLayer.n instead of src/MyLayer.js, and
src/MyLayer.js will be automatically generated. Only an underspecified
subset of ninjascript is implemented.

### DIFF
--- a/nin/backend/tests/TestLayer.n
+++ b/nin/backend/tests/TestLayer.n
@@ -1,0 +1,13 @@
+/* everything up until the first method is considered
+ * as the body of the constructor. */
+
+this.someMemberVariable = "hello world"
+
+someMethod(a, b, c) {
+    var scopedVariable = "someMethod was called"
+        log(scopedVariable, a, b, c)
+}
+
+someMethodThatReturnsSomething(a, b) {
+    < a * b
+}
--- a/nin/backend/ninjascript.js
+++ b/nin/backend/ninjascript.js
@@ -1,0 +1,54 @@
+function compile(moduleName, source) { 
+
+  function endsWith(str, suffix) {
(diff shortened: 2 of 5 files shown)


stianjensen:
Okay, now we're talking

nin/backend/tests/TestLayer.n
```diff
@@ -0,0 +1,13 @@
...
+
+this.someMemberVariable = "hello world"
+
+someMethod(a, b, c) {
+    var scopedVariable = "someMethod was called"
+        log(scopedVariable, a, b, c)
+}
+
+someMethodThatReturnsSomething(a, b) {
+    < a * b
```
stianjensen: Yeah, I don't like this syntax anymore :(

stianjensen:
Okay, so this is pretty neat.

Although, most of these syntax-slimming features are pretty redundant or minimal when ES6 class syntax comes into the mainstream. So I'm not sure if we actually want to complicate nin further by merging it.

sigvef:
> Okay, now we're talking
> 
> EDIT: and you've kept it a secret all this time!

I totally forgot that it even existed :P

nin/backend/tests/TestLayer.n
sigvef: Haha, me neither. Arbitrary cryptic symbols suck!

sigvef:
> Although, most of these syntax-slimming features are pretty redundant or minimal when ES6 class syntax comes into the mainstream. So I'm not sure if we actually want to complicate nin further by merging it.

I agree. Ninjascript doesn't actually bring much to the table. We should probably rather just start using es6-style classes and reevaluate the need for a new language later.

PR #214, opened 2016-05-15
Update and lock frontend npm dependencies

### LINKED ISSUE #205: Lock down frontend folder version numbers
All version numbers should be set to a specific version, otherwise Bad Things™ will probably happen.

stianjensen:
This fixes #205.

In confirming that things still work, I see that panel resizing (dragging the slider between the layers panel and the top panel, and between the left and right halves of the top panel) is broken. But as far as I can see that might have been broken before these changes, too.

Most dependencies have been updated to their newest version, as long as that worked. For grunt, updating to 1.0.0 has been held off because some of the dependencies we are using are not updated to be compatible with that grunt version yet.


sigvef:
> In confirming that things still work, I see that panel resizing (dragging the slider between the layers panel and the top panel, and between the left and right halves of the top panel) is broken. But as far as I can see that might have been broken before these changes, too.

Automatic resizing of the demo on panel resizing was disabled in c58e6e7866c73fb9b269601eccb781827cf79b49 and the divider between the top and bottom panels was fixed at 50% in 3466b583809cfd09c07b5a9d9e59900604b3f7c1, so there is no panel resizing regression in this patch afaict.

stianjensen:
Btw, the current version of grunt-contrib-imagemin (0.3.x) does not work for me. It fails upon installing pngquant.
It might be because of some weird changes on my system, seeing as it has not been broken before, but luckily newer versions of grunt-contrib-imagemin have stopped using pngquant, so this pull request fixes that issue.

PR #217, opened 2016-05-16
Normalize project root path

stianjensen:
Some commands (nin compile, for instance) depend on the projectPath
variable to parse file names.
There was an issue where filenames in FILES[] were written as
'es/filename.ext' (missing the r) when projectPath ended in a trailing
slash.

### DIFF
--- a/nin/backend/utils.js
+++ b/nin/backend/utils.js
@@ -10,10 +10,10 @@ function findProjectRoot(currentPath) {
         manifest = path.join(base, 'nin.json');
 
     if (fs.existsSync(manifest)) {
-      return base;
+      return base.replace(/\/$/, '');
     }
     up += '../';
-  } while (base != path.sep)
+  } while (base != path.sep);
 }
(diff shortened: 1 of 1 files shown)


nin/backend/utils.js
```diff
@@ -10,10 +10,10 @@ function findProjectRoot(currentPath) {
         manifest = path.join(base, 'nin.json');
 
     if (fs.existsSync(manifest)) {
-      return base;
+      return base.replace(/\/$/, '');
```
aleksanb: Is this portable across Filesystems? Might not fix the problem on Windows.

stianjensen: If that were the case, this shouldn't work either, I believe: https://github.com/ninjadev/nin/pull/217/files#diff-ff41c28559292673d9de3d8372834e69R15

aleksanb: You linked to this comment?

PR #218, opened 2016-05-16
Ensure crc input values are Buffer instances

### LINKED ISSUE #216: Nin compile broken
After testing old versions of nin, I've narrowed it down to this commit: 4016f96

Compiling with nin before that commit results in a working demo, but anything after that (remember to run `npm install`) will make the loading bar stop at 100% but no start button appear.
No errors in the console.

When not using png-compression the demo works fine.

System: OS X
Node: v6.1.0

stianjensen:
This fixes #216 

Since version 3.0.0, crc32 requires input values to be Buffer instances.

### DIFF
--- a/nin/backend/compress.js
+++ b/nin/backend/compress.js
@@ -18,7 +18,7 @@ function chunk(type, data) {
   return Buffer.concat([
       new Buffer(length, 'binary'),
       new Buffer(type + data, 'binary'),
-      new Buffer(positiveNumberToBytes(+('0x' + crc(type + data)), 4), 'binary')
+      new Buffer(positiveNumberToBytes(crc(new Buffer(type + data, 'binary')), 4), 'binary')
   ]);
 }
 
@@ -100,7 +100,7 @@ function compress(projectPath, payload, callback) {
   zlib.deflate(scanlinesBuffer.toString('binary'), function(err, buffer){
     var IDATData = Buffer.concat([
       buffer,
-      new Buffer(positiveNumberToBytes(+('0x' + crc(scanlinesBuffer.toString('binary'))), 4), 'binary')
+      new Buffer(positiveNumberToBytes(crc(scanlinesBuffer), 4), 'binary')
     ]);
     var IDATChunk = chunk('IDAT', IDATData.toString('binary'));
     callback(Buffer.concat([


sigvef:
Does this work on your machine? I get a crc mismatch when I try to compile a nin project with this branch.

Compiled with this branch:
image

Compiled with this branch, only with crc v2.0.0:
image

stianjensen:
> Does this work on your machine? I get a crc mismatch when I try to compile a nin project with this branch.

I get the same results as you using the same pngcheck-version.
But the demo works.

sigvef:
Perhaps we should downgrade crc to v2.x again. Thoughts?

stianjensen:
That will fix my compilation problems at least. Btw, is compilation also broken for you on v3.x?

sigvef:
Yes, it is broken for me on v3.x, but I didn't notice it until you just recently. Perhaps I hadn't actually run make in a while.

sigvef:
I tracked down to a change in behavior between crc 2.0.0 and crc 3.4.0. In 2.0.0, crc would treat input strings as binary byte arrays. In 3.4.0, crc treats the input strings as unicode strings and does an internal conversion which mangles the byte data for some input. The workaround is to make sure to always send a Buffer to crc. That means that instead of `crc(someString)`, we need to do `crc(new Buffer(someString, 'binary'))`.

nin/backend/compress.js
```diff
@@ -18,7 +18,7 @@ function chunk(type, data) {
   return Buffer.concat([
       new Buffer(length, 'binary'),
       new Buffer(type + data, 'binary'),
-      new Buffer(positiveNumberToBytes(+('0x' + crc(type + data)), 4), 'binary')
+      new Buffer(positiveNumberToBytes(+('0x' + crc(type + data).toString(16)), 4), 'binary')
```
sigvef: `new Buffer(positiveNumberToBytes(crc(new Buffer(type + data, 'binary')), 4), 'binary')`

nin/backend/compress.js
```diff
@@ -100,7 +100,7 @@ function compress(projectPath, payload, callback) {
   zlib.deflate(scanlinesBuffer.toString('binary'), function(err, buffer){
     var IDATData = Buffer.concat([
       buffer,
-      new Buffer(positiveNumberToBytes(+('0x' + crc(scanlinesBuffer.toString('binary'))), 4), 'binary')
+      new Buffer(positiveNumberToBytes(+('0x' + crc(scanlinesBuffer.toString('binary')).toString(16)), 4), 'binary')
```
sigvef: `new Buffer(positiveNumberToBytes(crc(scanlinesBuffer), 4), 'binary')`

sigvef:
The commit message could use an update if my other suggestions are taken.

stianjensen:
And we still need to convert to output to strings again, right?

stianjensen:
ping

PR #222, opened 2016-07-05
Make enter jump to cue point if available

sigvef:


aleksanb:
Does this fill the same need as forcing the playback to always be within the loop?
If so, would that feature be better?

sigvef:
I think there are valid use-cases for playing outside the loop. One example is if you want a specific loop, but are interested in watching a few seconds of "pre-roll" before entering the loop.

nin/frontend/app/scripts/directives/keybinding.js
```diff
@@ -32,7 +32,9 @@ angular.module('nin').directive('keybinding', function(commands, render, demo) {
     },
     '13': function(e) {
       // 'return'
-      commands.jumpToFrame(0);
+      var cuePoint = {};
+      commands.getCuePoint(cuePoint);
```
aleksanb: How about using a callback function instead?
It's a common pattern and then we won't have to use the proxy object!

sigvef:
ping

peterhgombos:
smooth.

sigvef:
Praise the lord

PR #227, opened 2016-07-12
Add nin update command

cristeahub:
This command will update nin itself.
This is especially useful for people like myself who aren't always up to date with nin!

It basically tries to do git pull in the nin folder, and if this is completed it will run the Makefile.

In the future we should do a more thorough check to see if the branch is up to date or not, and only run make if so, but right now this involved a lot of string matching, which seemed too hacky to me.


README.md
```diff
@@ -38,12 +38,14 @@ Run `make` in the nin folder. This is a convenience method that runs `make` in t
...
 
-To make nin available as a global command line utility, the easiest way is to add an alias to your `.bashrc` like this:
+To make nin available as a global command line utility, simply add the path to the backend to your `PATH` in your `.bashrc` file like this:
 
 ```
-alias nin='~/path/to/nin-project/backend/nin'
+export PATH="$PATH:/path/to/nin/nin/backend"
 ```
 
+If you are using Windows, please add this path to your `PATHEXT` variable.
```
iver56: My PATHEXT looks like this, and it doesn't look like I should add a path there
pathext

I'd rather use PATH

cristeahub: Ahh my bad, I misread the documentation from shelljs https://github.com/shelljs/shelljs#whichcommand . I'll change this :)

nin/backend/nin
```diff
@@ -63,6 +64,34 @@ program
...
+    if(!which('nin')) {
+      echo('Can\'t find nin executable in PATH/PATHEXT');
+      echo('Make sure \'/path/to/nin/nin/backend\' is present.');
+      exit(0);
+    }
+    echo('Checking for new version of nin');
+
+    var ninExecutable = which('nin');
+    var ninPath = ninExecutable.split('/');
+    var ninPath = ninPath.splice(0, ninPath.length-3).join('/');
```
aleksanb: You can remove everything above this line and replace it with `__dirname`,
as this points to the path of the executable folder.
See https://github.com/ninjadev/nin/blob/master/nin/backend/compress.js#L57 for an example.

cristeahub: Nice!

nin/backend/nin
```diff
@@ -63,6 +64,34 @@ program
...
+
+    cd(ninPath);
+    var gitGoToMaster = exec('git checkout master');
+    if(gitGoToMaster.code == 1) {
+      echo('Could not update nin, please resolve conflicts before updating');
+      exit(0);
+    }
+
+    var gitUpdate = exec('git pull');
```
aleksanb: Should check exit code. Changes in working space might make this command blow up.

nin/backend/nin
```diff
@@ -63,6 +64,34 @@ program
...
+    echo('Checking for new version of nin');
+
+    var ninExecutable = which('nin');
+    var ninPath = ninExecutable.split('/');
+    var ninPath = ninPath.splice(0, ninPath.length-3).join('/');
+
+    cd(ninPath);
+    var gitGoToMaster = exec('git checkout master');
+    if(gitGoToMaster.code == 1) {
+      echo('Could not update nin, please resolve conflicts before updating');
```
aleksanb: Should go to stderr.

nin/backend/nin
iver56: If you replace ninPath with __dirname, is there still a reason for having nin in PATH?

cristeahub: No, I removed that as well.

nin/backend/nin
```diff
@@ -63,6 +64,29 @@ program
...
 
+program
+  .command('update')
+  .description('Update nin to newest version')
+  .action(function() {
+    echo('Checking for new version of nin');
+
+    var ninExecutable = __dirname;
+    var ninPath = ninExecutable.split('/');
+    var ninPath = ninPath.splice(0, ninPath.length-3).join('/');
```
stianjensen: Redeclaring ninPath from the line above here

nin/backend/nin
cristeahub: The exit code is always 0 it seems, so the only solution here would be to check the messages produced, which I found very hacky, so I left it as an exercise to the....merger?

nin/backend/nin
cristeahub: This is intentional, the splice returns a list that needs to be saved.

cristeahub:
Ping!

nin/backend/nin
```diff
@@ -63,6 +64,29 @@ program
...
+
+    cd(ninPath);
+    var gitGoToMaster = exec('git checkout master');
+    if(gitGoToMaster.code == 1) {
```
aleksanb: Should check for nonzero, not 1.

nin/backend/nin
stianjensen: Okay, I'm trying to understand what the value of ninPath is intended to be.
`which('nin')` on my machine didn't return anything, since I had followed to install guide in the readme and only setup an alias in my .bashrc (nin was not in path).

`__dirname` returns "/Users/stiaje/Projects/nin", which I guess might be one less "nin" than what `which` returned.

`ninPath` is then finally set to "/Users" for me.

What is the value of ninPath when you run it?

cristeahub: Ok, we can't use `__dirname` in that case. On my computer it returns `/Users/cristea/Christoffersting/Projects/nin/nin/backend`. The code here normalizes the path by removing the last `nin` and `backend`.

aleksanb: This is the same behavior as me. Stian gon check again'

aleksanb: By the way, you just have to go up one level, and then you won't have to cd into nin again for that last ´make´ command.

stianjensen: Sorry, that path from me was wrong. It returns the path to the backend folder as you say.

cristeahub: > By the way, you just have to go up one level, and then you won't have to cd into nin again for that last ´make´ command.

Yeah, good catch!

cristeahub:
ping!

cristeahub:
ping

nin/backend/nin
```diff
@@ -63,6 +64,29 @@ program
...
+program
+  .command('update')
+  .description('Update nin to newest version')
+  .action(function() {
+    echo('Checking for new version of nin');
+
+    // normalize nin path to point to /path/to/nin-project/nin
+    var ninExecutable = __dirname;
+    var ninPath = ninExecutable.split('/');
+    var ninPath = ninPath.splice(0, ninPath.length-2).join('/');
```
stianjensen: the variable `ninPath` is already defined on the line above, so it shouldn't be redeclared with `var` here.

nin/backend/nin
```diff
@@ -63,6 +64,29 @@ program
...
+
+    cd(ninPath);
+    var gitGoToMaster = exec('git checkout master');
+    if(!(gitGoToMaster.code == 0)) {
```
stianjensen: Perhaps more idiomatic with `if (gitToToMaster.code != 0) {`

nin/backend/nin
```diff
@@ -1,15 +1,15 @@
 #!/usr/bin/env node
 
-var program = require('commander');
```
sigvef: This commit makes me happy :)

nin/backend/nin
```diff
@@ -63,6 +64,29 @@ program
...
 
+program
+  .command('update')
+  .description('Update nin to newest version')
+  .action(function() {
+    echo('Checking for new version of nin');
+
+    // normalize nin path to point to /path/to/nin-project/nin
+    var ninExecutable = __dirname;
+    var ninPath = ninExecutable.split('/');
```
sigvef: Consider using the `path` module for path operations.

nin/backend/nin
```diff
@@ -63,6 +64,29 @@ program
...
+    var ninPath = ninPath.splice(0, ninPath.length-2).join('/');
+
+    cd(ninPath);
```
sigvef: I guess these functions are magically imported via shelljs, but I was thrown off at first. Perhaps there is a way to import them in a more explicit fashion, so that one would use e.g. `shelljs.cd` or something?

nin/backend/nin
```diff
@@ -63,6 +64,29 @@ program
...
+    var gitGoToMaster = exec('git checkout master');
+    if(!(gitGoToMaster.code == 0)) {
+      console.error('Could not update nin, please resolve conflicts before updating');
+      exit(0);
```
sigvef: `process.exit(0)`? (or perhaps this is also from shelljs?)

nin/backend/nin
```diff
@@ -63,6 +64,29 @@ program
     generate.generate(projectRoot, resource, name);
   });
 
+program
+  .command('update')
+  .description('Update nin to newest version')
+  .action(function() {
+    echo('Checking for new version of nin');
```
sigvef: echo vs console.log

nin/backend/nin
cristeahub: Yes, from shelljs.

nin/backend/nin
cristeahub: This is from shelljs.

nin/backend/nin
cristeahub: I did include it as global in the top of the file, which is why these functions are here.

I can change it to locally import them, which will add the prefix.

nin/backend/nin
sigvef: Sure, but what is the advantage of using echo over just plain old console.log?

cristeahub: It's more idiomatic when you use shelljs, based on the example code.

stianjensen: I find it confusing to suddenly use echo here when we use console.log for the same thing everywhere else

cristeahub: I find it weird to use shelljs for this command, and then not use shelljs for some parts of the command.

aleksanb:
Needs slight amounts of cleanup before merge is possible.

cristeahub:
> Needs slight amounts of cleanup before merge is possible.

Yupp, first thing on the friday hackathon.

cristeahub:
ping

cristeahub:
Superseded by https://github.com/ninjadev/nin/pull/284 .

PR #233, opened 2016-07-13
Add checks for allowing playing outside of a loop

peterhgombos:
When setting a loop, currently it is impossible to play outside of this loop. This will only make the loop jump back when the demo is inside of the loop, and allowing for jumps outside of this loop.

### DIFF
--- a/nin/frontend/app/scripts/controllers/bottom.js
+++ b/nin/frontend/app/scripts/controllers/bottom.js
@@ -184,10 +184,14 @@
         $scope.loopEnd = $scope.loopStart + newLoopLength;
       });
 
+      var lastFrame = 0;
       $scope.$watch('currentFrame', function (nextFrame) {
-        if ($scope.loopActive && nextFrame >= $scope.loopEnd) {
+        if ($scope.loopActive && !$scope.demo.music.paused && lastFrame >= $scope.loopStart 
+                              && lastFrame <= $scope.loopEnd && nextFrame >= $scope.loopEnd 
+                              && !(nextFrame >= $scope.loopEnd + 5)) {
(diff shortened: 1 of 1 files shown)


nin/frontend/app/scripts/controllers/bottom.js
```diff
@@ -184,10 +184,14 @@
         $scope.loopEnd = $scope.loopStart + newLoopLength;
       });
 
+      var lastFrame = 0;
       $scope.$watch('currentFrame', function (nextFrame) {
-        if ($scope.loopActive && nextFrame >= $scope.loopEnd) {
+        if ($scope.loopActive && !$scope.demo.music.paused && lastFrame >= $scope.loopStart 
+                              && lastFrame <= $scope.loopEnd && nextFrame >= $scope.loopEnd 
+                              && !(nextFrame >= $scope.loopEnd + 5)) {
```
sigvef: Consider setting lastFrame = nextFrame somehow when demo.jumpToFrame is called -- that way you won't have to use this somewhat arbitrary guard to allow manually jumping out of loops.

aleksanb:
Wait what? Is just removing the loop not an alternative?

peterhgombos:
No, you might want to watch something outside of the loop without having to set the loop points all over again.

aleksanb:
How about changing the color of the loop segment when it is suspended?
That way we can easily see that it's paused!

peterhgombos:
It's not really suspended. The only thing that happens with this is that playback continues if you jump to another point outside of the loop.

aleksanb:
Needs rebase + addressing of sigves comment.

PR #248, opened 2016-07-17
Add metadata to compiled nin executables

aleksanb:
awesomee

:D

### DIFF
--- a/nin/backend/compile.js
+++ b/nin/backend/compile.js
@@ -1,6 +1,7 @@
 var chalk = require('chalk');
 var compress = require('./compress').compress;
 var exec = require('child_process').exec;
+var execSync = require('child_process').execSync;
 var fs = require('fs');
 var mkdirp = require('mkdirp');
 var p = require('path');
@@ -63,9 +64,21 @@ var compile = function(projectPath, options) {
         fs.writeFileSync(projectPath + '/bin/' + filename, data);
       });
     }
+
+    var settings = projectSettings.load(projectPath);
+    var metadata = [
+      ['Title', settings.title],
+      ['Author', settings.authors.join(', ')],
+      ['Creation time', '' + new Date()],
+      ['Software', 'NIN'],
+      ['Description',
+        "Git revision: " + execSync('git rev-parse HEAD') +
+        'Git origin: ' + execSync('git config --get remote.origin.url')]
+    ];
+
     if(options.pngCompress) {
       process.stdout.write(chalk.yellow('\nCompressing demo to .png.html'));
-      compress(projectPath, data, function(data) {
+      compress(projectPath, data, metadata, function(data) {
         renderOK();
         writeDemoToFile(data, 'demo.png.html');
         console.log(chalk.white('\n★ ---------------------------------------- ★'));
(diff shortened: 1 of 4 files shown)


nin/backend/compress.js
```diff
@@ -92,8 +94,14 @@ function compress(projectPath, payload, callback) {
           'demo=bootstrap({layers:layers, camerapaths:camerapaths, onprogress: ONPROGRESS, oncomplete: ONCOMPLETE});' +
         '}' +
       '}' +
-    '}</script><canvas class=hide height='+ height + ' width=' + width + '></canvas><img src=# onload=z()><!--';
+    '}</script>' +
```
sigvef: What does this change do?

nin/backend/compress.js
```diff
@@ -92,8 +94,14 @@ function compress(projectPath, payload, callback) {
...
+    '<canvas class=hide height='+ height + ' width=' + width + '></canvas><img src=# onload=z()><!--';
   var htMlChunk = chunk('htMl', new Buffer(html));
+  var metadataChunk = chunk('iTXt', new Buffer(
```
sigvef: `iTXt` chunks should probably follow the format in the spec, see http://www.libpng.org/pub/png/spec/1.2/PNG-Chunks.html (4.2.3.3. iTXt International textual data). You can use positiveNumberToBytes to get correct widths for the different fields.

aleksanb: Oh cool! I'll use all the appropriate fields.

nin/backend/compress.js
aleksanb: I found it easier to read on the next line.
Should i move it back?

sigvef: Ok, I wasn't sure if it was just a formatting thing or there was more meaning to it.

aleksanb:
Ping

nin/backend/compile.js
```diff
@@ -63,9 +64,21 @@ var compile = function(projectPath, options) {
         fs.writeFileSync(projectPath + '/bin/' + filename, data);
       });
     }
+
+    var settings = projectSettings.load(projectPath);
+    var metadata = [
```
sigvef: Consider:
```
var metadata = {
  Title: settings.title,
  Author: settings.authors.join(', '),
...
```

:)

sigvef: (unless they have to be in that order for some reason)

nin/backend/compress.js
```diff
@@ -92,7 +93,13 @@ function compress(projectPath, payload, callback) {
           'demo=bootstrap({layers:layers, camerapaths:camerapaths, onprogress: ONPROGRESS, oncomplete: ONCOMPLETE});' +
         '}' +
       '}' +
-    '}</script><canvas class=hide height='+ height + ' width=' + width + '></canvas><img src=# onload=z()><!--';
+    '}</script>' +
+    '<canvas class=hide height='+ height + ' width=' + width + '></canvas><img src=# onload=z()><!--';
+
+
+  var metadataChunks = Buffer.concat(metadata.map(function(data) {
+    return chunk('tEXt', new Buffer(data[0] + '\0' + data[1]));
```
sigvef: Perhaps use iTXt, since it supporst utf-8 instead of only latin-1.

nin/backend/compile.js
aleksanb: It's mainly in order for the metadata tEXt to look nicer with the title first. When generating the OG tags I think I might switch to a dictionary, as i need to reference the authors field directly then.

nin/backend/compress.js
aleksanb: Turns out implementing iTXt support is rather difficult, and finding a tool that's able to show those segments even more so.
I suggest we wait until a need for it appears.

PR #249, opened 2016-07-17
Always use project index.html

aleksanb:
We now bundle the new-and-improved index.html from ninjadev/dix as the default index.html.
The pouet link is left pointing to the Ninjadev page, i think this is a reasonable default.

This removes one of the last should-be-demo specific parts from nin proper! Rejoice!


nin/backend/blank-project/index.html
```diff
@@ -0,0 +1,130 @@
...
+    loadingText.classList.add('hide');
+    progressBarContainer.classList.add('hide');
+    startButton.classList.remove('hide');
+  };
+
+  STARTDEMO = function() {
+    wrapper.classList.remove('visible');
+    wrapper.style.background = '#fff';
+    document.body.style.background = '#000';
+    demo.start();
```
sigvef: Perhaps we need to setTimeout demo.start in order to let the browser repaint before performing warmup.

cristeahub: Would be a good idea in my opinion. Ended up doing this on Crankwork in order to let the button change name to `Starting..!`, since the warmup does take a couple of seconds.

You can change this function to something like this (taken directly from Crankwork):
```
STARTDEMO = function() {
  startButton.innerHTML = 'Starting..!';
  setTimeout(function() {
    wrapper.classList.remove('visible');
    wrapper.style.background = '#fff';
    document.body.style.background = '#000';
    demo.start();
  }, 100);
};
```

aleksanb:
Updated to newest version from dix.
This also includes og tags which can be filled out by the user.
In the future, these tags can be autogenerated by nin.

aleksanb:
Superseded by new PR

PR #252, opened 2016-07-18
Metadata improvements

aleksanb:
Wow such improve much great.
Creating tooling is such fun!


nin/backend/blank-project/index.html
```diff
@@ -0,0 +1,144 @@
...
+    startButton.innerHTML = 'Starting..!';
+    setTimeout(function() {
+      wrapper.classList.remove('visible');
+      wrapper.style.background = '#fff';
+      document.body.style.background = '#000';
+      demo.start();
+    }, 100);
+  };
+  </script>
+  <!--<script>
```
sigvef: Commented script?

aleksanb: I was unsure whether you want google analytics bundled by default, and in that case what analytics ID you should use.
I can make it a default option, and then template in a GA key from nin.json.

sigvef: That sounds good.

aleksanb:
Ping. All operational. Tested against newest dix with both https://github.com/ninjadev/dix/pull/139 and https://github.com/ninjadev/dix/pull/138 merged, and both the .html and .html.png have correct metadata now and run in the browser fine.

PR #254, opened 2016-07-20
Shift paradigm from layers to nodes

sigvef:
image

This commit shifts the central paradigm of nin from being a layer-based
system to being a node-based system. Not all features have been ported
over to the new system, although it is certainly complete enough to be
used to make demos. Hopefully we are willing to merge this sooner rather
than later, and we can then later fix the rest of the features as they
are needed.

* Overview

Before, a demo in nin would consist of several layers in an ordered
list. Each layer would have a start time and end time, and each
layer's update and render methods would be called with the usual
update/render call semantics while demo time was within a layer's start
and end times. Each layer exposed an EffectComposer pass, and all passes
for active layers were chained together in a top-to-bottom such that the
next layer could optionally access the rendered data of the previous
layer.

Now, a demo in nin consists of several nodes in a graph. Nodes do not
have start times or end times. Rather, the graph is traversed from a
designated root node in order call update and render for each of the
reachable nodes. A node may enable or disable their inputs, which serves
as temporary graph modification in order to "connect and disconnect"
different nodes at different times in the demo. This is the spiritual
replacement of start times and end times from the layer system.  The big
advantage of the new node-based system over the old layer system  is
that it is now easier to create more advanced multi-pass rendering
pipelines.

* Nodes and the graph

Nodes have different types, and all nodes inherit from the base
NIN.Node. A node type may declare a number of named and typed inputs and
outputs. Nodes may produce data and store it in their outputs. Connected
nodes may then read that same data from their inputs. The directed graph
of connections from node outputs to node inputs may not contain cycles.

The graph is stored in res/graph.json, which is analogous to the old
res/layers.json. Because of the frequency of graph edits in the UI
(particularly dragging nodes to change their position), and in the
interest of simplicity and performance, res/graph.json is not actively
watched for changes on disk after the first load at the beginning of
`nin run`. The intention is that most changes to res/graph.json should
be made via the UI.

Currently, all graphs *must* have a node of type NIN.RootNode called
root in order for the demo to update and render properly. This node is
the root node that is used for traversal during update and render, and
the texture made available to the root node's only input root.screen is
what is finally showed on the screen.

The LayerManager of old has been replaced with an analogous NodeManager.

* Current state of features

This section elaborates upon the different features of nin and their
current state with regards to the paradigm shift. Features that are not
affected are probably not mentioned here.

** Node graph viewer / editor

Replacing the old layer viewer/editor, the node graph viewer / editor
currently supports visualization of the graph including connections and
activeness of nodes. Nodes are persistently draggable. The viewer is
pan/zoom-enabled, but has no pan/zoom persistence. It is currently not
possible to create or delete nodes or connections.

** Generate

This has not been touched, so it most likely does not work.

** Compilation

This has not been touched, so it most likely does not work.

** Init

This has not been touched, so it most likely does not work.

** Backend res/layers.json / res/graph.js manipulation

This has been updated and works for whatever it is being used for in
this commit, but has not been scrutinized for completeness.

** Shader updates

This is probably broken. Perhaps we can leverage the node system in
order to make it less grep-y in the future.

** Camera paths, fly-around mode

This is probably broken.

** Layer / Node inspection

Nodes can no longer be inspected, but inputs and outputs can. Currently,
the only inspectable inputs/outputs are texture input/outputs. When
inspected, they will render their texture to the demo screen after the
regular render pass is finished.


sigvef:
See https://github.com/sigvef/nin-node-test-demo for a project that works with this prq.

iver56:
How about backwards compatibility with existing demos? Should we simply create a [release](https://github.com/ninjadev/nin/releases) before this prq is merged, and use that release for old demos?

sigvef:
This will not work with any old demos at all. Just like most other changes to nin ;)

aleksanb:
Currently each layer is wrapped by a closure, and the layer is stored on window such as:
```
global.SceneSwitcherNode = SceneSwitcherNode;
})(this);
```

If we use es6 modules or something similar, or wrap the nodes outside the file, we can reduce the amount of boilerplate!

aleksanb:
Yeah, let's tag current head with LEGACY and merge this into master when it has been reviewed.

iver56:
Tag created:
https://github.com/ninjadev/nin/releases/tag/v0.1.0

sigvef:
> If we use es6 modules or something similar, or wrap the nodes outside the file, we can reduce the amount of boilerplate!

I agree, that would be nice. From what I'm reading, using proper es6 modules require a smarter packer/assembler than the current concat-and-maybe-optimize scheme we're using right now. Do you have a concrete example of how it could be done for us without resorting to changing the build flow? I'm incidentally also concerned about 'use strict' used without closures.

One way we could do this is is something like:
```
(function(NIN) {
  NIN.SomethingNode = class extends NIN.Node {
    ..
  }
})(this.NIN);
```
I'm not sure if this is semantically any different from what we have now, so it might just come down to a matter of preference.

Changing the build flow down the line might be useful, but I'm not sure it should be a priority for now.

nin/backend/graph.js
```diff
@@ -0,0 +1,68 @@
...
+
+function update(projectPath, nodeId, data, callback) {
+  read(projectPath, function (err, graph) {
+    if (err) {
+      if (callback) {
+        callback(err);
+      }
+      return;
+    }
+    var index;
```
aleksanb: Equivalent to `var index = array.find(node => node.id == nodeId)`

nin/backend/socket.js
```diff
@@ -39,8 +39,10 @@ function socket(projectPath, onConnectionCallback) {
       switch (event.type) {
         case 'set':
           var data = {};
-          data[event.data.field] = event.data.value;
-          layers.update(projectPath, event.data.id, data, function (err) {
+          for(var key in event.data.fields) {
```
aleksanb: Man i wish github could let me highlight regions for commenting.
Anyways, isn't this part equivalent to just passing event.data.fields directly to graph?

nin/backend/watch.js
```diff
@@ -29,12 +29,21 @@ function watch(projectPath, cb) {
...
 
+  var graphAlreadyLoaded = false;
+
   watcher.on('all', function (event, path) {
     if (event === 'unlink') event = 'delete';
     if (event == 'addDir') {
       return;
     }
 
+    if(path == 'res/graph.json') {
```
aleksanb: Should we add an issue explaining this change, and linking to it from the code?
It's mentioned in the PR (that editing mainly will happen from UI) but its bound to bite someone in the ass at one point when they can't figure out why their graph.json isn't pushed over the intertubes.

nin/dasBoot/Node.js
```diff
@@ -0,0 +1,22 @@
...
+  'use strict';
+
+  class Node {
+    constructor(id, options) {
+      this.id = id;
+      this.inputs = options.inputs || {};
+      this.outputs = options.outputs || {};
+    }
+
+    resize() {
```
aleksanb: Is there such a thing as forcing a subclass to implement functions from it's parent?
This would be a good fit for abstract / virtual.

sigvef: One strategy is to throw in the base class. For non-vital methods, however, I'd rather they stay optional in case a node doesn't have anything to do in that method.

nin/dasBoot/NodeManager.js
```diff
@@ -0,0 +1,132 @@
...
+    this.graphChangeListeners = [];
+  }
+
+  onGraphChange(fn) {
+    this.graphChangeListeners.push(fn);
+  }
+
+  fireGraphChange() {
+    console.log('graph change!');
+    for(var i = 0; i < this.graphChangeListeners.length; i++) {
```
aleksanb: I'm just a sucker for map now that => has been introduced!
`this.graphChangeListeners.map(gcl => gcl())`

nin/dasBoot/NodeManager.js
```diff
@@ -0,0 +1,132 @@
...
+    }
+    this.nodes[node.id] = node;
+    this.fireGraphChange();
+  }
+
+  connect(fromNodeId, outputName, toNodeId, inputName) {
+    this.nodes[fromNodeId].outputs[outputName].destination =
+      this.nodes[toNodeId].inputs[inputName];
+    this.nodes[toNodeId].inputs[inputName].source =
+      this.nodes[fromNodeId].outputs[outputName];
```
aleksanb: Do connections have to be reflexive?
Should a child node know about its parents?

nin/dasBoot/NodeManager.js
```diff
@@ -0,0 +1,132 @@
...
+    console.log('graph change!');
+    for(var i = 0; i < this.graphChangeListeners.length; i++) {
+      this.graphChangeListeners[i]();
+    }
+  }
+
+  createNode(nodeInfo) {
+    nodeInfo.options = nodeInfo.options || {};
+    var node;
+    if(nodeInfo.type.slice(0, 4) == 'NIN.') {
```
aleksanb: Can we check for subclassing of NIN.Node instead?
This seems a bit brittle.

nin/dasBoot/NodeManager.js
```diff
@@ -0,0 +1,132 @@
...
+      node = new NIN[nodeInfo.type.slice(4)](nodeInfo.id, nodeInfo.options);
+    } else {
```
aleksanb: What is the case for nodes that don't inherit from NIN.Node?

nin/dasBoot/NodeManager.js
```diff
@@ -0,0 +1,132 @@
...
+    this.nodes[fromNodeId].outputs[outputName].destination =
+      this.nodes[toNodeId].inputs[inputName];
+    this.nodes[toNodeId].inputs[inputName].source =
+      this.nodes[fromNodeId].outputs[outputName];
+    this.fireGraphChange();
+  }
+
+  resize() {
+    for(var key in this.nodes) {
+      this.nodes[key].resize();
```
aleksanb: Can the order of resizes result in weird glitches?
If a parent node is resized before the child node, using a not-yet resized child node result in its own computation.
In that case a postorder traversal would be best.

nin/dasBoot/NodeManager.js
```diff
@@ -0,0 +1,132 @@
...
+    this.fireGraphChange();
+  }
+
+  resize() {
+    for(var key in this.nodes) {
+      this.nodes[key].resize(); 
+    }
+  }
+
+  traverseNodeGraphPostOrderDfs(node, fn, visitedSet) {
```
aleksanb: Ah--the function is already ready!

nin/dasBoot/NodeManager.js
```diff
@@ -0,0 +1,132 @@
...
+            fn,
+            visitedSet);
+      }
+    }
+    fn(node);
+  }
+
+  update(frame) {
+    if(!this.nodes.root) {
+      return;
```
aleksanb: If there's always supposed to be a root-perhaps throw instead?

nin/dasBoot/NodeManager.js
aleksanb: By making visitedSet a default parameter the callee doesn't have to care about it.
That is `traverseNodeGraphPostOrderDfs(node, fn, visitedSet={}) {`.

nin/dasBoot/NodeManager.js
```diff
@@ -0,0 +1,132 @@
...
+    }
+    this.traverseNodeGraphPostOrderDfs(this.nodes.root, function(node) {
+      node._graphEditorInfo.active = true;
+      node.update(frame);
+    }, {});
+    for(var key in this.nodes) {
+      if(this.nodes[key]._graphEditorInfo.oldActive !=
+          this.nodes[key]._graphEditorInfo.active) {
+        console.log('changed active state', key, this.nodes[key], this.nodes[key]._graphEditorInfo.oldActive, this.nodes[key]._graphEditorInfo.active);
+        this.fireGraphChange();
```
aleksanb: Does fireGraphChange have to fire for each changed node, or is one time for changes sufficient?
If there are performance-heavy callbacks it's prob better to do it just once.

nin/dasBoot/NodeManager.js
```diff
@@ -0,0 +1,132 @@
...
+    if(!this.nodes.root) {
+      return;
+    }
+    renderer.clear(true, true, true);
+    this.traverseNodeGraphPostOrderDfs(this.nodes.root, function(node) {
+      node.render(renderer);
+    }, {});
+  }
+
+  reset() {
```
aleksanb: Stubbing for later implementation?

nin/dasBoot/NodeManager.js
```diff
@@ -0,0 +1,132 @@
...
+      }
+    }
+  }
+
+  render(renderer) {
+    if(!this.nodes.root) {
+      return;
+    }
+    renderer.clear(true, true, true);
+    this.traverseNodeGraphPostOrderDfs(this.nodes.root, function(node) {
```
aleksanb: This is where the magic happens.
Is renderer the THREE.JS renderer? My eyes hurt from the triple anonymous boolean parameters ^^.

nin/dasBoot/NodeManager.js
sigvef: No, nodeInfo.type is a string.

nin/dasBoot/RootNode.js
```diff
@@ -0,0 +1,34 @@
...
+
+      this.camera = new THREE.OrthographicCamera(-1, 1, 1, -1, 0, 1); 
+      this.scene = new THREE.Scene();
+      this.quad = new THREE.Mesh(new THREE.PlaneBufferGeometry(2, 2), null);
+      this.scene.add(this.quad);
+      this.material = new THREE.MeshBasicMaterial();
+    }
+
+    render(renderer) {
+      if(renderer.overrideToScreenTexture) {
```
aleksanb: Is this where we add preview-code for the different nodes?

nin/dasBoot/RootNode.js
```diff
@@ -0,0 +1,34 @@
...
+        this.quad.material = new THREE.MeshBasicMaterial({
+          map: renderer.overrideToScreenTexture
```
aleksanb: Did we always render things to a quad in front of the camera?

nin/dasBoot/NodeManager.js
sigvef: Doubly-linked makes everything much easier. I can't think of a concrete example where it is needed on the spot, but I think it is a good idea.

nin/dasBoot/ShaderPassNode.js
```diff
@@ -0,0 +1,36 @@
...
+          A: new NIN.TextureOutput()
+        }
+      });
+      this.camera = new THREE.OrthographicCamera(-1, 1, 1, -1, 0, 1); 
+      this.scene = new THREE.Scene();
+      this.quad = new THREE.Mesh(new THREE.PlaneBufferGeometry(2, 2),
+                                 new THREE.ShaderMaterial(options.shader));
+      this.shader = this.quad.material;
+      this.scene.add(this.quad);
+      this.renderTarget = new THREE.WebGLRenderTarget(640, 360, {
```
aleksanb: Does this make all shaders render at a fixed 640x360?

nin/dasBoot/ShaderPassNode.js
```diff
@@ -0,0 +1,36 @@
...
+        minFilter: THREE.LinearFilter, 
+        magFilter: THREE.LinearFilter, 
+        format: THREE.RGBFormat
+      });
+    }
+    
+    resize() {
+      this.renderTarget.setSize(16 * GU, 9 * GU);
```
aleksanb: But it seems to be overridden here. Is the previous value just a default then?

nin/dasBoot/ShaderPassNode.js
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    render(renderer) {
+      renderer.render(this.scene, this.camera, this.renderTarget, true);
+      this.outputs.A.setValue(this.renderTarget.texture);
```
aleksanb: A since it is the first output?

nin/dasBoot/NodeManager.js
sigvef: The check isn't for whether or not they inherit from NIN.Node. All nodes inherit from NIN.Node. The check is for whether or not the class is defined in the NIN namespace. An example of a node class that is defined outside of the NIN namespace is a node type that is defined in a demo project, e.g. MyAwesomeRotatingCubeSceneNode.

nin/dasBoot/NodeManager.js
sigvef: Good question. Resizes should be graph-independent, as far as I can tell. Really the only use for resize is for a node to resize it's internal canvases and rendertargets, both of which do not rely on inputs.

nin/dasBoot/TextureInput.js
```diff
@@ -0,0 +1,8 @@
+(function(NIN) {
+  'use strict';
+
+  class TextureInput extends NIN.Input {
```
aleksanb: Is this subclassed just for type safety?
Then again JS doesn't check types.

nin/dasBoot/NodeManager.js
sigvef: Nice, didn't know that was part of es6. Does this have the same problem as default arguments in Python, i.e. that all invocations of the function share the same literal object as their visitedset?

nin/dasBoot/NodeManager.js
sigvef: Because of the way nodes are loaded, there is a brief window of time right after page load where no nodes have been added. If we throw instead of silently ignore, things will break.

nin/dasBoot/NodeManager.js
sigvef: Once is sufficient. The callbacks are somewhat performance-intensive, which is why the next line is break-statement, ensuring that fireGraphChange only gets called once :)

nin/frontend/app/scripts/controllers/main.js
```diff
@@ -255,23 +265,34 @@
...
               $scope.fileCache[event.path] = event.content;
+              var indirectEval = eval;
               indirectEval(event.content);
-              layerShaderDependencies[event.layername] = event.shaderDependencies;
 
-              demo.lm.refresh(event.layername);
-              demo.lm.update(demo.looper.currentFrame);
+              var splitted = event.path.split('/');
+              var filename = splitted[splitted.length - 1];
+              var typename = filename.slice(0, -3);
```
aleksanb: "File extensions exactly 3 characters long ought to be enough for anyone".

nin/dasBoot/NodeManager.js
sigvef: The renderer is indeed the THREEJS renderer. The triple unnamed boolean parameters are for clearing three different kinds of buffers.

nin/dasBoot/NodeManager.js
```diff
@@ -0,0 +1,132 @@
...
+    for(var key in node.inputs) {
+      node.inputs[key].node = node;
+    }
+    for(var key in node.outputs) {
+      node.outputs[key].node = node;
+    }
+    return node;
+  }
+
+  insertNode(node) {
```
aleksanb: ReplaceNode might be more accurate?
I when i encountered it further down (in the WS node push reload code) i got confused by the insertion.

nin/dasBoot/NodeManager.js
sigvef: Yes, the rest of these stubs are just included to avoid crashes, as they are called by various parts of the system. With the new paradigm, some of them might not be needed either. Time will tell. For now, it seemed like an ok idea to just stub them.

nin/dasBoot/RootNode.js
sigvef: Maybe not all, since perhaps not all nodes should be previewed by being rendered by the THREEJS renderer. I chose this as a hacky entrypoint because it was easy for textures.

nin/dasBoot/RootNode.js
sigvef: Yes and no. Just rendering a plain ol' THREE.Scene directly to the screen does not require rendering a quad. Chaining renderings with the EffectComposer in the old layer paradigm would render all but the last pass into a rendertarget so that it could be fed as a texture into the next pass. For shaderpasses and copypasses this quad technique was used. In effect, this last pass of the rendering chain in the new system is a copypass, since it renders an incoming texture right to the screen.

nin/dasBoot/ShaderPassNode.js
sigvef: Yes, the previous value is just a default. It needs to be inited to *something*, so I chose the current default size of the preview window in the nin GUI.

aleksanb:
I couldn't find relativeFrame / frame in the nodes.
Seems like nodes only receive the global frame counter now?
This would make animating scenes with our current workflow rather tricky,
having to handle subtraction of the nodes startFrame within the node itsel?

nin/dasBoot/ShaderPassNode.js
sigvef: A is the name of the output. It could be something semantically useful like 'depth' or 'color' something like that. Since this is a generic node, I went with A because it has no particular connotations in this setting other than being first in the alphabet and is coincidentally short enough to not cause display problems in the graph editor.

aleksanb:
Could you make a tracking issue for all major features missing,
so we can update it during further development?

nin/dasBoot/TextureInput.js
sigvef: It is subclassed just for type safety. JS doesn't often check types, but we can. This lets us for instance display different UI widgets for different types of inputs down the line, if we want.

aleksanb:
Have you thought about pros/cons of a flat/nested graph.json?

nin/dasBoot/NodeManager.js
sigvef: insertOrReplaceNode() ?

upsertNode() ?

nin/dasBoot/NodeManager.js
aleksanb: A concrete example would be nice.
Otherwise i'm for isolating children from their parents as much as possible.

nin/dasBoot/NodeManager.js
aleksanb: Nope! They're evaluated at each call time!

sigvef:
> I couldn't find relativeFrame / frame in the nodes.
> Seems like nodes only receive the global frame counter now?
> This would make animating scenes with our current workflow rather tricky,
> having to handle subtraction of the nodes startFrame within the node itsel?

Since nodes no longer have start times and end times relativeFrame doesn't really make sense. I think a good solution might be to let nodes manipulate the time that children see. That way a SceneSelectorNode can effectively supply a modified frame in addition to an absolute frame. Perhaps we can call this modified relativeFrame, but really it doesn't mean exactly the same.

absoluteFrame and frame? frame and modifiedFrame? frame and relativeFrame?

sigvef:
> Have you thought about pros/cons of a flat/nested graph.json?
> I guess moving connections around is easier in the flat version.

Having a flat graph.json is much easier to work with in my opinion.

sigvef:
> Could you make a tracking issue for all major features missing,
> so we can update it during further development?

Perhaps we should make separate issues for broken features when we notice that they are broken?

nin/backend/watch.js
sigvef: I'm not sure if an issue is the right tool, but we can add a prominent warning the README, perhaps.

aleksanb:
> I couldn't find relativeFrame / frame in the nodes.
>     Seems like nodes only receive the global frame counter now?
>     This would make animating scenes with our current workflow rather tricky,
>     having to handle subtraction of the nodes startFrame within the node itsel?
> 
> Since nodes no longer have start times and end times relativeFrame doesn't really make sense. I think a good solution might be to let nodes manipulate the time that children see. That way a SceneSelectorNode can effectively supply a modified frame in addition to an absolute frame. Perhaps we can call this modified relativeFrame, but really it doesn't mean exactly the same.
> 
> absoluteFrame and frame? frame and modifiedFrame? frame and relativeFrame?

Frame is fine and i agree the sceneselectornode can handle it.
The question becomes if the sceneselectornode should be promoted to something special for UX purposes, superseding the role of the layer playlist.

As the sceneselector gon' get huge for our projects, showing only one of its children would be nice.

aleksanb:
I'm done. Now for @stianjensen s comments.

Everything looks good, main point of annoyance is boilerplates for all nodes,
not sure what the best solution is.

nin/dasBoot/NodeManager.js
sigvef: One concrete example is when replacing an existing node. In order to reconnect outputs it is convenient to use outputs.somename.destination to see where the output should be connected.

nin/dasBoot/ShaderPassNode.js
sigvef: No, for reference:

image

nin/frontend/app/scripts/controllers/main.js
sigvef: Exactly 2, actually, since the third character is `.`. Luckily, since this only handles .js-files, it should be ok.

sigvef:
> As the sceneselector gon' get huge for our projects, showing only one of its children would be nice.

I'm not sure what you mean by showing only one of its children, but it is possible to preview the inputs for a node by selecting it.

nin/backend/graph.js
sigvef: array.findIndex(..), but otherwise nice catch!

nin/backend/socket.js
sigvef: Indeed it is.

nin/dasBoot/NodeManager.js
sigvef: :D

sigvef:
ping

nin/dasBoot/NodeManager.js
aleksanb: InsertOrReplace's fine.

sigvef: InsertOrReplace it is.

aleksanb:
issue

I am only able to do `nin run` from the root folder of the project.
Something isn't being done relative to projectRoot.

aleksanb:
clicking

Clicking on the sceneselect node crashes the backend.

aleksanb:
undefinedrt

When playing / pausing / resizing the window one frequently gets TypeError: this.renderTarget is undefined
 which sometimes breaks playback.

aleksanb:
Turns out clicking or dragging anything in the frontend crashes the backend with:

/home/aleksanb/Projects/nin/nin/backend/socket.js:41
          graph.update(projectPath, event.data.id, data.fields, function (err) {
                                                   ^

ReferenceError: data is not defined
    at SockJSConnection.<anonymous> (/home/aleksanb/Projects/nin/nin/backend/socket.js:41:52)
    at emitOne (events.js:77:13)

I am using Firefox.

sigvef:
> I am only able to do nin run from the root folder of the project.
> Something isn't being done relative to projectRoot.

This is broken in master as well, so should be fixed in a separate pull request. I made an issue: https://github.com/ninjadev/nin/issues/255 .

sigvef:
Regarding the crashes when interacting with the graph editor, that was caused by a typo I introduced when rewriting some things that were pointed out in the last round of code review. I've fixed it now.

sigvef:
> When playing / pausing / resizing the window one frequently gets TypeError: this.renderTarget is undefined
> which sometimes breaks playback.

I'm not seeing this. Are you still seeing this after my latest rebase?

aleksanb:
Try with Firefox. At an art museum currently.

aleksanb:
Start resizing right after the page starts loading.

sigvef:
As it turns out, the resizing error was rooted in a bug in demo project itself ( https://github.com/sigvef/nin-node-test-demo ), not nin. I've fixed the bug there.

aleksanb:
likedis

Not sure if this is a bug with the demo project or something else, but i never see Lena when i run the demo.

sigvef:
Lena shows up at frame < 1000, and the baboon at frame >= 1000.

aleksanb:
> Lena shows up at frame < 1000, and the baboon at frame >= 1000.

For me, i just get no animation when frame < 1000 (with white + baboon), and at >= 1000, they both start spinning.

aleksanb:
Okay i found the issue, if you start playing before all the resources have loaded, everything breaks.
If i wait untill all loading is finished, lena works.

sigvef:
Hm, on my machine hitting play before everything has loaded works fine (Chrome).

README.md
```diff
@@ -9,6 +9,14 @@ nin is Ninjadev's internal demo tool. It is a tool for easing development of bro
 This project has a node backend that keeps track of all files and compiles files as they are edited.
 The frontend of this project is written in Angular and displays among other the layers that the demo consists of.
 
+
+## nin is now node-based!
+
+Nin just became node-based instead of layer-based.
+Here is a list of gotchas to watch out for if you are used to layer-based nin:
+
+- Changing res/graph.json on does *not* trigger updates in a running nin instance.
```
stianjensen: typo: remove the 'on'

sigvef:
ping

aleksanb:
I know that Stian wants to do a full review of the codebase before merge, but had a bit spotty internet in Bodø.

aleksanb:
In Regard to how we review commits of this size, I think it might be useful to do commits that address comments as separate commits that we then squash via the web panel at merge time. It might be easier to catch regressions that way, and for single-commit PRs like this it shouldn't introduce too much overhead.

sigvef:
I agree, I will pile on additional commits for each extra change I make from now on. For reference, my last update was removing the word 'on', as per https://github.com/ninjadev/nin/pull/254#discussion_r71971158 .

cristeahub:
I too would like to do a full review of this before this is merged. Such a major change will benefit from multiple people looking at it. We can halt creating new features for nin to make this change easier.

stianjensen:
The layer zooming seems to be broken.
Zooming in works, but zooming out does not.

From my testing, it seems to be caused the fact that the horizontal scrollbar is now caused by overflow on the body element, and the calculations for xScale does not work anymore.
Setting the .bottom element to `width:100%;overflow-x:scroll` and hardcoding the width of .layers-bar-container to the same width as the waveform fixes this issue for me.

This also has the effect that when scrolling in the bottom timeline, the whole body element is not scrolled.

stianjensen:
What are your thoughts on how demos should be structured into scenes or separate parts with this new system?

In the demo project, there is a SceneSwitcherNode. Would we create one big SceneSwitcherNode for each demo which deals with enabling and disabling different nodes, or would nin provide some sort of builtin, more declarative, system for dealing with this?

sigvef:
On time/scene structure in demos: I opted for a hands-off approach in nin that leaves scene switching largely in the hands of the demo developer. Nin supports enabling and disabling of node inputs in order to effectively connect and disconnect subtrees from the graph, but does not lay down any conventions on how and when these node inputs should be enabled or disabled. One way of handling scene switching would be to have a multiplexing node with hard-coded timestamps (like SceneSwitcherNode). Another would be to have a multiplexer that triggers scene switches by other exotic inputs such as midi, aunino, or other things. Ultimately, I expect a convention for scene switching will emerge across future demos made with nin. At that point we can consider bringing a more opinionated scene switcher into nin.

sigvef:
On horizontal scrolling in time: perhaps we don't even need horizontal zoom any more? Perhaps a fully zoomed-out full-width view is enough.

stianjensen:
> On horizontal scrolling in time: perhaps we don't even need horizontal zoom any more? Perhaps a fully zoomed-out full-width view is enough.

That's true. It's probably better to just remove it.

aleksanb:
I think we should be pragmatic and make a default implementation when it comes to scene switching.
It's such a core feature that it'll be a shame if everyone has to roll their own.
If everyone has to make their own sceneswitcher node we're going to end up with a fragmented mess and higher threshold to make new projects .

cristeahub:
> I think we should be pragmatic and make a default implementation when it comes to scene switching.
It's such a core feature that it'll be a shame if everyone has to roll their own.

I agree here. At least one way to do it now with good support as a starting point. Should some better feature emerge, then we can always change the default scene switcher, or have multiple.

sigvef:
Ok, how do you propose the default scene switcher should work? As a multiplexer node presumably, but with how many inputs? How should switch timestamps be specified?  What about transitions between scenes, e.g. cross-fades, wipes, etc work (node-based nin finally makes these things easy to implement)? What about cutting back and forth between scenes? These are questions that we need to answer before we can make a default scene switcher. Answering them probably requires some testing and experimentation to see what works and what doesn't, which is why I'm suggesting that we defer adding a default scene switcher until we know more about how we would want a default scene switcher to work. That is, let's make some demos, and then use what we've learned from those demos to choose a default scene switcher for nin.

sigvef:
To be clear: I *do* want a default scene switcher, but I don't think we should add it right now.

cristeahub:
I've started the review here, and will continue tomorrow. Will hold off my feedback until tomorrow, as I think I might need to revisit some of my thoughts 👍

cristeahub:
I like the change, it does seem like a way to do a lot of what we want easier, however I do have some thoughts:

- **The graphs will become very big and incomprehensible**: A solution here could be to hide parts of the graph in a vitual supernode or something. I.e. one scene might consist of 20 nodes and they can collapse to 1 virtual node that just says scene 1, when clicked it can show the 20 nodes it's made up of.

- **We can't easily see how time affects the demo**: We lose a sense of what happens where with this new system. I have no good idea here yet, and this might be fixed by our down-the-road scene system. As of right now it seems almost to me that we could have our old scene layout, but with each scene being its own graph. Would like input here!

- **This breaks a lot**: Since this a big breaking change we will have to rewrite a lot of nin. This isn't an issue, but we just need to be fully aware of this. We can't expect to create a fully fledged demo with this system until a lot more is in place. We have the layer-based release (https://github.com/ninjadev/nin/releases/tag/v0.1.0) so we can always fall back on this.

nin/backend/serve.js
```diff
@@ -92,8 +90,18 @@ var serve = function(projectPath, shouldRunHeadlessly) {
...
+        var directoryAScore = directoryPrecedence[a.slice(0, 3)];
+        var directoryBScore = directoryPrecedence[b.slice(0, 3)];
+        if(directoryAScore == directoryBScore) {
+          return a > b;
+        }
+        return directoryAScore > directoryBScore;
+
+      });
       for (var i in sortedPaths) {
+        console.log('first we need one of these!', sortedPaths[i]);
```
cristeahub: This message seems a bit out of place. Maybe change the text or consider removing it. I guess it was for debug purposes?
image

nin/dasBoot/NodeManager.js
cristeahub: > One concrete example is when replacing an existing node.

I suspect this will be a common occurence with having a demo as a big graph, so I'm all for the reflexivity here.

nin/dasBoot/NodeManager.js
cristeahub: What's the details here on the loading part? Seems like we should change the way the loading works (I guess this is a nice issue for after this prq is merged).

nin/dasBoot/NodeManager.js
cristeahub: I suggest just adding this for clarification `renderer.clear(true, true, true); // clear all three different buffers`. This will be easier if js introduces argument labels, but as of today this is not a thing in es6 to my knowledge.

nin/dasBoot/RootNode.js
cristeahub: Where do we add the preview-code for different nodes?

nin/dasBoot/NodeManager.js
sigvef: In broad strokes, loading works like this:

1. The nin angular app is loaded
2. graph.json is fetched and parsed, and nodes are new'ed
3. everything is groovy

This is similar to how layers.json is loaded in layer-based nin.

nin/dasBoot/RootNode.js
sigvef: I haven't given it much thought. Perhaps we can wait and see what kind of demands different node previews have before we decide where we put preview code.

sigvef:
ping

sigvef:
(I haven't fixed horizontal zooming, but perhaps we can defer that to a later pull request).

stianjensen:
I'm gonna go ahead and get this merged now, before the pull request gets even longer and harder to follow. @cristeahub brought up some nice points in his summary, but I think all of those points can be addressed in future changes.

Can you squash your commits Sigve, and then I'll get this into master?

sigvef:
ping

stianjensen:
🍫 🍰  🍸 🎉

cristeahub:
:cheer:

stianjensen:
:missing_emoji:

aleksanb:
Any progress on this?

sigvef:
Yes: https://github.com/ninjadev/nin/compare/782a977a841f3b47dbb14442029de1322b537788...master

aleksanb:
Great.

PR #260, opened 2016-08-20
Use projectPath when reading files from disk

### LINKED ISSUE #255: `nin run` does not work from subfolders of a nin project
backend/watch.js does not use projectPath, so `nin run` crashes if the working directory is not the project root.

stianjensen:
This fixes #255.

### DIFF
--- a/nin/backend/serve.js
+++ b/nin/backend/serve.js
@@ -57,7 +57,7 @@ var serve = function(projectPath, shouldRunHeadlessly) {
       var path = data.path,
           filename = p.basename(path),
           filenameWithoutExtension = filename.split('.')[0],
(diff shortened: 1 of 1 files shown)


nin/backend/serve.js
```diff
@@ -57,7 +57,7 @@ var serve = function(projectPath, shouldRunHeadlessly) {
       var path = data.path,
           filename = p.basename(path),
           filenameWithoutExtension = filename.split('.')[0],
-          content = fs.readFileSync(path, 'utf-8');
+          content = fs.readFileSync(projectPath + '/' + path, 'utf-8');
```
sigvef: Consider using path.join.

stianjensen:
ping

PR #261, opened 2016-08-20
Disable horizontal zooming

### LINKED ISSUE #226: In fly-around mode left and right arrow should not scroll the timeline

### LINKED ISSUE #258: Horizontal zooming
With #254, horizontal zooming is broken. Considering the current feature set of node-based nin, horizontal zoom doesn't seem very useful, so it might be best to simply remove the feature altogether.

On the other hand, new timeline-based features might be introduced later, making zooming useful again. Still, we can always just create a new zoom-system when it's needed (since the current one is broken anyway).

I suggest we simply remove it for now.

stianjensen:
The music timeline now always fills the width of the screen.

This fixes #258.
This fixes #226.

### DIFF
--- a/nin/frontend/app/scripts/controllers/bottom.js
+++ b/nin/frontend/app/scripts/controllers/bottom.js
@@ -5,9 +5,6 @@
     .controller('BottomCtrl', function ($scope, $interval, socket, camera, commands) {
 
       $scope.xScale = 1;
-      $scope.xScaleTarget = 1;
-      $scope.yScale = 1;
-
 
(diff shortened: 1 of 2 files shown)


sigvef:
The music waveform disappears on resize:

image

stianjensen:
> The music waveform disappears on resize:

Not on my machine :(
Testing more.

stianjensen:
Can you do an extra hard mega-refresh just to be sure? Having trouble reproducing here.

sigvef:
Yup, ultra mega hard refresh did the trick.

PR #263, opened 2016-08-20
Node-based Camera controller

stianjensen:
Check out https://github.com/stianjensen/nin-cam-controller-test for an example of how it's used


stianjensen:
This should be complete and functioning now, including live reloading.

stianjensen:
Fly-around is still missing, but I'll tackle that later.

sigvef:
make no longer works

image

stianjensen:
ping

nin/dasBoot/CameraNode.js
```diff
@@ -0,0 +1,31 @@
...
+        outputs: {
+          camera: new NIN.CameraOutput()
+        }
+      });
+
+      this.options = options;
+
+      this.cameraController = new CameraController();
+      this.outputs.camera.setValue(this.cameraController.camera);
+      Loader.loadAjax(options.path, response => this.initializeCamera(response));
```
aleksanb: Eta reduction with `(options.path, this.initializeCamera)`?

aleksanb: How does this loader work together with compilation?
As there won't be a file to load from then?

stianjensen: Like any other call to loadAjax I suppose

stianjensen: > Eta reduction with (options.path, this.initializeCamera)?

What do you mean?

aleksanb: You can remove the lambda.

PR #267, opened 2016-11-08
Switch Grunt to webpack

stianjensen:
This is instead of migrating to a anew major version of grunt and trying
to update all of its deprecated plugins.
In the process, all bower dependencies are also moved over to npm so now
we only need one package manager.

In the process, all files needed to be modularized.
When doing that, all controllers and services were migrated to ES6 class
syntax.
For ES6 class syntax compatibility, Angular's controllerAs pattern is also employed.

A few of the big code changes are mostly white-space so remember to add ?w=1 when reviewing.


nin/frontend/app/scripts/controllers/top.js
```diff
@@ -1,62 +1,54 @@
-angular.module('nin').controller('TopCtrl', function($scope, camera, commands, $interval) {
-  $scope.displayValue = function(id, val) {
+class TopCtrl {
+  constructor($scope, camera, commands) {
+    console.log('topmute', $scope.mute, 'yolo', this.mute, 'mega', $scope.main.mute);
```
stianjensen: console.log

stianjensen:
Locked package.json dependencies now

PR #279, opened 2016-11-11
Make music bar fill entire width

### LINKED ISSUE #272: Music bar is not wide enough
It should go all the way out to the edge:

image

sigvef:
This also fixes an issue where the body had a small useless scrollbar.

This fixes #272.


sigvef:
image

PR #285, opened 2016-11-11
Robustify

aleksanb:
This PR significantly reduces the number of unnecessary exceptions printed when certain errors occur in NIN.

### DIFF
--- a/nin/frontend/app/scripts/controllers/main.js
+++ b/nin/frontend/app/scripts/controllers/main.js
@@ -205,19 +205,16 @@ class MainCtrl {
         case 'graph':
           let graph = JSON.parse(event.content);
 
-          this.graph = graph;
-          demo.nm.hardReset();
-
           for (let nodeInfo of graph) {
+            let node = demo.nm.createNode(nodeInfo);
+
             try {
-              let node = demo.nm.createNode(nodeInfo);
               demo.nm.insertOrReplaceNode(node);
             } catch (e) {
               // This hack only works due to not-yet received
               // nodes created through generate not having
               // any connections / friends.
               $timeout(function () {
-                let node = demo.nm.createNode(nodeInfo);
                 demo.nm.insertOrReplaceNode(node);
               }, 100);
             }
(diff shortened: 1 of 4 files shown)


nin/frontend/app/scripts/controllers/main.js
```diff
@@ -205,19 +205,16 @@ class MainCtrl {
         case 'graph':
           let graph = JSON.parse(event.content);
 
-          this.graph = graph;
-          demo.nm.hardReset();
```
sigvef: hardReset no longer needed?

nin/frontend/app/scripts/filters/errorExcerpt.js
```diff
@@ -1,5 +1,9 @@
 function errorExcerpt($sce) {
   return function(error, fileCache) {
+    if (fileCache === undefined || !error.path in fileCache) {
+      return;
```
sigvef: Does `error` just disappear when we do this? Shoud we perhaps throw the error or something to compensate?

nin/frontend/app/scripts/controllers/main.js
aleksanb: HardReset never did anything :O
https://github.com/ninjadev/nin/blob/550b6726a45c739954e31c030c2928c329a0a282/nin/dasBoot/NodeManager.js#L122

nin/frontend/app/scripts/filters/errorExcerpt.js
aleksanb: Our error handling code is actual spaghetti atm:
https://github.com/ninjadev/nin/blob/550b6726a45c739954e31c030c2928c329a0a282/nin/frontend/app/views/top-panel.html#L20

Three things are always printed in the error display:
- The first line of the errors
- The file excerpt where the errror came from (This is not always appliccable, as the error might not have a file source, as is the case when the error originates from our GUI for instance)
- The rest of the lines of the errors.

This makes it so that the file excerpt doesn't crash NIN while trying to show the original error,
while still showing the full error.

PR #286, opened 2016-11-11
Eslint backend

cristeahub:
This prq lints the backend.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,10 @@ The frontend of this project is written in Angular and displays among other the 
 
 To install nin simply run the command `npm install -g ninjadev-nin`.
 
+## Requirements
+
+This projects requires node version `7.1.0` or newer.
+
 ## nin is now node-based!
 
 Nin just became node-based instead of layer-based.
--- a/nin/backend/package.json
+++ b/nin/backend/package.json
@@ -8,6 +8,7 @@
     "commander": "2.9.0",
     "concat-files": "0.1.0",
     "crc": "3.4.0",
+    "eslint": "3.10.0",
     "express": "4.13.4",
     "fluent-ffmpeg": "2.0.1",
     "glob": "7.0.3",
@@ -18,8 +19,9 @@
     "walk": "2.3.9"
   },
   "scripts": {
-    "npm_prepublish": "sed -i.bak s/'var publishing = false;'/'var publishing = true;'/g serve.js && make bundle-deps",
-    "npm_postpublish": "sed -i.bak s/'var publishing = true;'/'var publishing = false;'/g serve.js"
+    "lint": "eslint *.js",
+    "npm_prepublish": "sed -i.bak s/'const publishing = false;'/'const publishing = true;'/g serve.js && make bundle-deps",
+    "npm_postpublish": "sed -i.bak s/'const publishing = true;'/'const publishing = false;'/g serve.js"
   },
   "bin": {
     "nin": "./nin"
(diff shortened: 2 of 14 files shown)


nin/backend/package.json
```diff
@@ -18,6 +19,7 @@
     "walk": "2.3.9"
   },
   "scripts": {
+    "lint": "./node_modules/.bin/eslint *.js",
```
aleksanb: No need to prefix with node modules in the scripts section :dancer:

cristeahub:
ping

sigvef:
image

sigvef:
Can't check on newer versions right now, but perhaps we should add something to the README about the node versions supported.

cristeahub:
ping

README.md
```diff
@@ -13,6 +13,10 @@ The frontend of this project is written in Angular and displays among other the
 
 To install nin simply run the command `npm install -g ninjadev-nin`.
 
+## Requirements
+
+This projects requires node version `7.1.0` or newer.
```
stianjensen: Do we want to be this strict? I mean, it's fine by me, since it's already in homebrew, but it's only released 5 days ago so other platforms might not get it in packaging anytime soon.

Node 6.9, however is LTS and might be a better baseline. https://nodejs.org/en/blog/release/v6.9.0/

nin/backend/utils.js
```diff
@@ -50,10 +51,10 @@ function mergeOptions(input, defaults) {
 function colorizeCommanderHelpText(txt) {
   return (txt.replace(/(\[[^\]]*\])/g, chalk.grey('$1'))
              .replace(/Usage: (\w+)/g, 'Usage: ' + chalk.green('$1'))
-             .replace(/(\n  \w+ [^\n]*)/g, chalk.cyan('$1'))
-             .replace(/(\n   ( -+[\w,-]+)+)/g, chalk.green('$1'))
-             .replace(/(\n    \w+)/g, chalk.green('$1'))
-             .replace(/(\n  \w+:)/g, chalk.yellow('$1')));
+             .replace(/(\n{2}\w+ [^\n]*)/g, chalk.cyan('$1'))
```
sigvef: Did eslint suggest these changes? They are not equivalent. Before:

image

After:

image

cristeahub: Yes this was eslint. Can change

README.md
aleksanb: All of us on ubuntu are on <= 4.2.6 atm :(

aleksanb:
It appears that changing to node 6.9 will break a lot of our buffer-related code as functions have been changed.

README.md
sigvef: I just installed from source, now I have v8.0.0-pre.

cristeahub:
> It appears that changing to node 6.9 will break a lot of our buffer-related code as functions have been changed.

I guess 7.1.0 is best then

aleksanb:
What i mean is that anything over >=6.9 has changed a lot of the buffer related code, which means that compilation won't work. there either.
Right now compilation doesn't work anyways, so that might be okay for now.
7.1 also has even more sweet es6 features, so that might be okay when we're first breaking stuff anyways.

stianjensen:
Difference between 6.9 and 7.1 ES6 wise is very small I think. But we should update our Buffer code to support 6.9

aleksanb:
I just installed 7.1 from nodejs.org. It was very simple, and easy to verify the checksum.

cristeahub:
ping! I let the spaces remain to make sure we won't get any issues with cross-terminal compatability.

cristeahub:
> I just installed 7.1 from nodejs.org. It was very simple, and easy to verify the checksum.

Seems we should go for 7.1 in my opinion.

PR #287, opened 2016-11-12
Repair nin init

### LINKED ISSUE #278: nin init does not work
It broke when we moved to a node-based system.

sigvef:
This fixes #278.


nin/backend/generate/generate.js
```diff
@@ -19,10 +19,14 @@ var generate = function(projectRoot, type, name) {
...
 
       graph.transform(projectRoot, function(g) {
-        g.push({
+        const object = {
           id: name,
-          type: camelizedName
-        });
+          type: camelizedName,
+        };
+        for(const key in options) {
```
aleksanb: You can replace this with the new (and sexy) `Object.assign(object, options)` 🍾

stianjensen: Can't you also use spread syntax inside the object literal?

aleksanb: Yeah, that might work as well!
What a time to be alive!

sigvef:
> Can't you also use spread syntax inside the object literal?

That's what I was trying to pull off on Friday, to no avail. Can you show how that would work? If not I'll go with Object.assign.

stianjensen:
> That's what I was trying to pull off on Friday, to no avail. Can you show how that would work? If not I'll go with Object.assign.

Did you have an updated node version then?

sigvef:
Nope, but I tried the same tricks after upgrading.

stianjensen:
Use Object.assign, object spread syntax is still only a stage 3 proposal it turns out

sigvef:
ping

PR #288, opened 2016-11-12
Restructure NIN project

aleksanb:
This restructuring simplifies the publishing process as everything is located under a single package.json as opposed to the previous structure of two package.jsons on the same level, requiring us to copy files over for publishing.


sigvef:
We're not going with the project structure that we discussed?

```
nin/
├── backend/
│   ├── package.json
├── dasBoot/
├── frontend/
│   ├── package.json
└── package.json
```

aleksanb:
I tried that at first, but with the rationale that i didn't want to rewrite all references to __dirname as i thought __dirname always was the directory of the executable ran, i unrolled all files under nin/.
As it turns out that __dirname is module-local (https://nodejs.org/docs/latest/api/globals.html#globals_dirname) i can switch to the discussed structure instead.

aleksanb:
Updated to reflect the all-new all-old structure!

nin/backend/serve.js
```diff
@@ -10,17 +10,9 @@ var readDir = require('readdir');
 var socket = require('./socket');
 var watch = require('./watch');
 
-var publishing = false;
-
 
 var serve = function(projectPath, shouldRunHeadlessly) {
-
-  var genPath = p.join(projectPath, '/gen/');
-  mkdirp.sync(genPath);
```
sigvef: This used to create the `gen/` folder in a demo project's folder. We probably still want to do this, as other parts of nin assume that this directory exists.

aleksanb:
Addressed, it worked locally due to me having created gen from an earlier run of nin.

stianjensen:
How does this work when published? npm handles installing dependencies from packge.json in nested directories?

aleksanb:
No idea how it works when published.

stianjensen:
I guess it would work if the dependencies listed in backend/package.json are moved out to the top-most package.json

aleksanb:
I'll do that then.

aleksanb:
It's done.
Publish through make publish, as that will build the frontend project so that it can be added during the publishing step.

PR #294, opened 2016-11-13
Lint remaining backend files

stianjensen:
The lint command in package.json is updated so that the eslint runs with
backend as cwd. This ensures that it loads the correct .eslintignore

Templates in the generate folder are ignored since they're not strictly
part of the backend, but templates for user projects.

Also, one lint error remains, but it's there because the feature is
still broken so I figured it serves as a nice reminder to fix the
feature.

### DIFF
--- a/nin/backend/compile.js
+++ b/nin/backend/compile.js
@@ -1,52 +1,52 @@
-var chalk = require('chalk');
-var compress = require('./compress').compress;
-var exec = require('child_process').exec;
-var execSync = require('child_process').execSync;
-var fs = require('fs');
-var mkdirp = require('mkdirp');
-var p = require('path');
-var projectSettings = require('./projectSettings');
-var rmdir = require('rimraf');
-var shaderGen = require('./shadergen').shaderGen;
-var utils = require('./utils');
-var walk = require('walk');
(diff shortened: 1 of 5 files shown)


nin/backend/compile.js
```diff
@@ -82,26 +82,26 @@ var compile = function(projectPath, options) {
...
-        .join('\n');
+      .map(key => `<!-- ${key}: ${metadata[key]} -->`)
+      .join('\n');
 
     const ogTags =
-`<meta property="og:title" content="${utils.unsafeHTMLEscape(metadata.Title)}" />
-<meta property="og:description" content="${utils.unsafeHTMLEscape(metadata.Description)}" />
-<meta property="og:image" content="${metadata.previewImage}" />
-<meta name="author" content="${utils.unsafeHTMLEscape(metadata.Author)}" />`;
+      `<meta property="og:title" content="${utils.unsafeHTMLEscape(metadata.Title)}" />
```
sigvef: This changes the contents of the string itself. Intentional?

stianjensen: Done automatically by eslint. Lemme take a look.

PR #295, opened 2016-11-13
Add travis.yml

stianjensen:
This is working now. Not the cleanest, prettiest config, but it should be working.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,10 @@
+language: node_js
+node_js:
+  - "node"
+install:
+  - npm install -g yarn
+  - yarn
+  - cd nin/frontend && yarn && cd -
+script:
+  - cd nin/backend && ../../node_modules/.bin/eslint .
+  - cd ../frontend && node_modules/.bin/eslint app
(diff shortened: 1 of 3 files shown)


stianjensen:
`|| true` problems:

screenshot 2016-11-13 03 25 33

sigvef:
Hm yes I expected as much :/

cristeahub:
What was the reason for `|| true`?

sigvef:
It was added because `npm run lint` adds a bunch of crap to the output if the script exits with a non-zero return code. `|| true` was a hack to make sure that the linting script always exited nicely, even when there were errors.

Without the hack:

image

With the hack:

image

cristeahub:
Ahh right! Yeah, I noticed that too. I guess we can have Travis specific instructions.

aleksanb:
So... do we merge this?

stianjensen:
No, it's a WIP. I will ping when I've updated.

stianjensen:
There's still one lint error in the backend, but it's an actual error because we've stopped using layers.

sigvef:
Can you add a quick fix for that as a second commit in this pull request? That way we won't break the build in our other outstanding pull requests if we merge this.

stianjensen:
ping

PR #296, opened 2016-11-13
Lock up the .ninrc, clean up the menubar shortcuts

aleksanb:
This commit drains the menubar shortcuts (or 'swamp').
Previously the swamp was spread across multiple files.

In addition, this commit 'locks up' user preferences in the newly introduced '.ninrc' file,
opening for endless possibilities for user customization.
One can now bind up-arrow and down-arrow to black and white themes, respectively, allowing for the greatest switching of all time.


aleksanb:
wtf. 
image

stianjensen:
There's no travis.yml on this branch, so travis is probably just guessing what to do. I tried disabling travis checking now.

aleksanb:
Travis is still completely nuts :(

sigvef:
What happens when we want to have keybindings that are not also a menu item? One example is 'x' to resetFlyFlightDynamics, which this pull request deletes completely.

sigvef:
Perhaps the keyboard shortcuts should be linked to commands directly, and then the menu mapping can get it's shortcut help text from the commands module.

aleksanb:
I reccommend not having any hidden keybindings,
as it really hinders discoverability. Better to expose everything, and use submenus if you're afraid of too large menus.

The removing of resetFlyFlightDynamics is accidental, although without functioning cameraFLyaroundController at the moment, so nobody will notice :P

aleksanb:
I went for a centralized solution with one file for all keybinding and menu related stuff to reduce the mental overhead of having to look everywhere for the source of the commands.

aleksanb:
It isn't strictly normalized then, but sometimes denormalized data is nice.

sigvef:
> I reccommend not having any hidden keybindings,
> as it really hinders discoverability. Better to expose everything, and use submenus if you're afraid of too large menus.

I agree (except if we start getting more modal keybindings for things like flying the camera etc).

> The removing of resetFlyFlightDynamics is accidental, although without functioning cameraFLyaroundController at the moment, so nobody will notice :P

Might as well remove it.

> I went for a centralized solution with one file for all keybinding and menu related stuff to reduce the mental overhead of having to look everywhere for the source of the commands.

Yes, centralized is good, this is better than what we have now.

> Perhaps the keyboard shortcuts should be linked to commands directly, and then the menu mapping can get it's shortcut help text from the commands module.

What do you think about this suggestion?

aleksanb:
Much has been improved now, read the new PR text for updates.

nin/frontend/app/scripts/services/ninrc.js
```diff
@@ -0,0 +1,9 @@
+function ninrc($http) {
+  const xhr = new XMLHttpRequest();
```
stianjensen: `fetch` totally has more than good enough browser support for use in nin now, btw.

https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch

EDIT: Looking at this again, I realize it is implemented synchronously. If it need to be sync, I guess it's best to just leave it like it is.

sigvef: Yeah, it needs to be sync. Perhaps we should have specified that in a source comment.

README.md
```diff
@@ -83,42 +83,20 @@ The only rules you need to oblige is that:
...
+Currently, keyboard shortcuts is the only behavior which can be changed in the .ninrc.
+The list of canonical names for keybindings can be found in
+[nin/frontend/app/scripts/directives/menubar.js](https://github.com/ninjadev/nin/blob/master/nin/frontend/app/scripts/directives/menubar.js).
+
+An example .ninrc looks like the following:
+
+```
+[keybinds]
+startRendering=left
+stopRendering=right
```
sigvef: These shortcut names are sent directly to keymage, so whatever works there works for us https://github.com/piranha/keymage .

stianjensen:
When testing, I get the following error on load:
```
angular.js:13920TypeError: Cannot read property 'undefined' of undefined
    at http://localhost:8000/app.bundle.js:57391:41
    at Array.forEach (native)
    at http://localhost:8000/app.bundle.js:57390:25
    at Array.forEach (native)
    at Object.link (http://localhost:8000/app.bundle.js:57389:13)
    at http://localhost:8000/app.bundle.js:11549:19
    at invokeLinkFn (http://localhost:8000/app.bundle.js:20236:10)
    at nodeLinkFn (http://localhost:8000/app.bundle.js:19637:12)
    at http://localhost:8000/app.bundle.js:19975:14
    at processQueue (http://localhost:8000/app.bundle.js:26685:29) <div menubar="" menu="main.menu" class="ng-scope">
```

aleksanb:
Ping.

nin/frontend/app/scripts/services/ninrc.js
stianjensen: screenshot 2016-11-28 22 23 19

:(

PR #297, opened 2016-11-13
Make nin init take directory, init as git repository

aleksanb:
Modelled after rust's cargo, nin init now takes a required directory
argument, avoiding accidentally initializing nin projects in your
current directory (as many zip-users may be familiar with).

Additionally, nin projects now come with git initialized as well as a
bundled .gitignore tailored for our project format.
This will reduce onboarding friction for new users, as well as
profilerating the spread of git and allowing for easier sharing on
github.

### DIFF
--- a/nin/backend/generate/generate.js
+++ b/nin/backend/generate/generate.js
@@ -27,7 +27,7 @@ const generate = function(projectRoot, type, name, options) {
         if (err) {
           console.error(err);
         } else {
-          console.log('Added ' + name + ' to graph.json');
+          console.log(`-> added ${name} to graph.json`);
         }
       });
       break;
@@ -53,7 +53,7 @@ const generate = function(projectRoot, type, name, options) {
         if (err) {
           console.error(err);
         } else {
-          console.log('Added ' + shaderLayerName + ' to layers.json');
+          console.log(`-> added ${shaderLayerName} to layers.json`);
         }
       });
       break;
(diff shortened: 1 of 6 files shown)


nin/backend/init.js
```diff
@@ -1,19 +1,28 @@
...
+    process.exit(1);
   }
+  child_process.execSync(
+      `git init ${projectPath}`,
+      {stdio: 'inherit'});
+
   glob(path.join(__dirname, 'blank-project/*'), function(error, files) {
+    // As globs don't expand to include dotfiles,
+    // we need to add the .gitignore manually
+    files.push(path.join(__dirname, 'blank-project/.gitignore'));
```
sigvef: Perhaps globbing for 'blank-project/.*' works? Then we won't have to remember to special case for other dotfiles that we want in the future as well.

nin/backend/generate/generate.js
```diff
@@ -53,7 +53,7 @@ const generate = function(projectRoot, type, name, options) {
         if (err) {
           console.error(err);
         } else {
-          console.log('Added ' + shaderLayerName + ' to layers.json');
+          console.log(`-> added ${shaderLayerName} to layers.json`);
```
sigvef: layers.json no longer exists (but this could be fixed in a different pull request).

nin/backend/init.js
```diff
@@ -32,7 +32,7 @@ function init(dirname) {
           }
         });
         projectSettings.init(projectPath);
-        console.log(chalk.green('This directory is now a nin project. Do'),
+        console.log(chalk.green(`${projectPath} is now a nin project. Run`),
```
sigvef: Perhaps this message should specify in which directory to run `nin run`?

aleksanb:
I agree with the views expressed here.
Fixup coming up.

aleksanb:
Updated PR.
It's now become an omnibus bill, and README improvements have sneaked themselves in as well (both documenting nin new as well as the rest of the file).

PR #298, opened 2016-11-14
Potentially fix race-condition regression with layer generation

aleksanb:
Layers -> Generate node in the frontend seems to crash while it didn't
before. If i remember correctly, this is one of the reasons createNode
was inside the try-catch to begin with, and is therefore moved back.

### DIFF
--- a/nin/frontend/app/scripts/controllers/main.js
+++ b/nin/frontend/app/scripts/controllers/main.js
@@ -206,17 +206,17 @@ class MainCtrl {
           let graph = JSON.parse(event.content);
 
           for (let nodeInfo of graph) {
-            let node = demo.nm.createNode(nodeInfo);
-
             try {
+              let node = demo.nm.createNode(nodeInfo);
               demo.nm.insertOrReplaceNode(node);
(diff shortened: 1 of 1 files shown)


nin/frontend/app/scripts/controllers/main.js
```diff
@@ -206,17 +206,17 @@ class MainCtrl {
...
               demo.nm.insertOrReplaceNode(node);
             } catch (e) {
               // This hack only works due to not-yet received
               // nodes created through generate not having
               // any connections / friends.
               $timeout(function () {
+                let node = demo.nm.createNode(nodeInfo);
                 demo.nm.insertOrReplaceNode(node);
-              }, 100);
+              }, 200);
```
sigvef: What is the reasoning behind increasing the timeout (not that it matters, just curious)?

aleksanb: Just re-moving the createnode line inside the try-catch didn't seem to help, but i can't remember if increasing the number to 200 helped either. Might need more testing.

PR #300, opened 2016-12-26
Replace closure compiler with js implementation

sigvef:
The Google Closure Compiler now has an official js port, so we can use
that instead. The upside for us is that we no longer have a Java
dependency, and we can have a cleaner-looking compile implementation
that doesn't rely on external executables.

Of course, compilation is currently broken in master since the switch to
nodes instead of layers, so compiling anything on this commit won't
technically work. However, this should "totally work" when we fix the
other stuff.

### DIFF
--- a/nin/backend/compile.js
+++ b/nin/backend/compile.js
@@ -1,6 +1,7 @@
 const chalk = require('chalk');
+const glob = require('glob');
+const closureCompiler = require('google-closure-compiler-js');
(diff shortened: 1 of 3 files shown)


aleksanb:
Could you add your changes to the yarn.lock as well?

sigvef:
ping

aleksanb:
Needs rebase.

sigvef:
ping

PR #301, opened 2016-12-26
Add optipng compression of res/*.png

sigvef:
image


nin/backend/compile.js
```diff
@@ -43,12 +47,34 @@ function res(projectPath, callback) {
     root += '/';
     root = root.replace(/\/\//g, '/');
 
-    const file = fs.readFileSync(root + stat.name);
+    let file = fs.readFileSync(root + stat.name);
```
aleksanb: Nitpick: file doesn't seem to be redeclared in scope?

aleksanb:
Could you add your changes to the yarn.lock as well?

nin/backend/compile.js
sigvef: I changed it for a first implementation that redeclared `file`, but then I forgot to change it back when I changed to a different implementation that didn't redeclare it. Fixed.

sigvef:
ping

PR #305, opened 2016-12-27
Disable Travis email notifications

sigvef:


stianjensen:
Huh, I weren't getting those :/

sigvef:
The default is to send emails to the committer, more or less.

PR #307, opened 2017-01-14
Upgrade dependencies

sigvef:
Security updates and bug fixes \o/


stianjensen:
Uhm. how did strings end up being non-single quote on this branch only?

Also, is that why this was closed?

cristeahub:
> Also, is that why this was closed?

because of this ^^ (was a game on ninjacon)
https://github.com/ninjadev/nin/pull/307/files#diff-8ee2343978836a779dc9f8d6b794c3b2R2417

stianjensen:
haha, that was actually my first though when I saw the pull request (that it was Sigve trying to sneak something into yarn.lock). But I was on mobile, so didn't scroll through all of it.

PR #308, opened 2017-01-15
Refactor nodeManager instantiation logic

aleksanb:
Nonexisting constructors (due to race conditions) and errors in your
node's constructor are now two separate errors again, to everyones
delight!

### DIFF
--- a/nin/dasBoot/NodeManager.js
+++ b/nin/dasBoot/NodeManager.js
@@ -7,19 +7,16 @@ class NodeManager {
 
   createNode(nodeInfo) {
     nodeInfo.options = nodeInfo.options || {};
-    var node;
-    try {
-      if(nodeInfo.type.slice(0, 4) == 'NIN.') {
-        node = new NIN[nodeInfo.type.slice(4)](nodeInfo.id, nodeInfo.options);
-      } else {
-        node = new window[nodeInfo.type](nodeInfo.id, nodeInfo.options);
-      }
-    } catch (e) {
-      const errorMessage = "NodeManager: Attempted to instantiate nonexisting node type: " + nodeInfo.type;
-      console.error(errorMessage);
-      throw new Error(errorMessage);
+
+    const nodeConstructor = nodeInfo.type.slice(0, 4) === 'NIN.' ?
+      NIN[nodeInfo.type.slice(4)]
+      : window[nodeInfo.type];
+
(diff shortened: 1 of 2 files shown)


nin/frontend/app/scripts/controllers/main.js
```diff
@@ -78,7 +78,8 @@ class MainCtrl {
             try {
               let node = demo.nm.createNode(nodeInfo);
               demo.nm.insertOrReplaceNode(node);
-            } catch (e) {
+            } catch(e) {
+              console.error(e);
```
sigvef: Will this flood the logs with the same error 5 times a second? Should we differentiate between "node hasn't been loaded yet" vs "node crashed when we tried to init it"?

aleksanb: We should, yes. Currently the code is a spaghetti ball of yarn, and needs a refactor to be able to properly differentiate between the two cases.

aleksanb:
Updated to comply with the NIN2017 release.

nin/dasBoot/NodeManager.js
```diff
@@ -7,19 +7,16 @@ class NodeManager {
...
-        node = new NIN[nodeInfo.type.slice(4)](nodeInfo.id, nodeInfo.options);
-      } else {
-        node = new window[nodeInfo.type](nodeInfo.id, nodeInfo.options);
-      }
-    } catch (e) {
-      const errorMessage = "NodeManager: Attempted to instantiate nonexisting node type: " + nodeInfo.type;
-      console.error(errorMessage);
-      throw new Error(errorMessage);
+
+    const constructor = nodeInfo.type.slice(0, 4) === 'NIN.' ?
```
stianjensen: Using constructor as a variable name _might_ work, but it also has meaning in js, so could we use something else here?

aleksanb: How about `const radicalCat = ...` ?

aleksanb: `const utroTorsk = ...`?

stianjensen: I was thinking something more in the lines of `nodeConstructor` 🤔

aleksanb: `const nodeConstructor = ...`?

aleksanb:
Ping.

Issue #276, opened 2016-11-11
ES6ify the code base


cristeahub:
The code base (especially the backend) hasn't been updated to es6 yet.

Update the following to es6:
- [ ] Frontend
- [ ] Backend

stianjensen:
Frontend has started the migration slowly. Controllers have been updated to ES6 class syntax etc.

stianjensen:
For dasBoot, we will probably want to hold off on updating and consider the performance impact of updating syntax.

cristeahub:
> performance impact of updating syntax.

Is there a big impact there?

stianjensen:
Depends.

https://kpdecker.github.io/six-speed/

PR #313, opened 2017-02-04
Modernize js syntax in nin executable

sigvef:


cristeahub:
I guess this ties up to https://github.com/ninjadev/nin/issues/276 . Any way we way we can close that issue with this prq?

sigvef:
There are other parts that haven't been upgraded yet, though, so we could just leave the issue open until we are happy with the entire codebase.

PR #316, opened 2017-02-05
Add stats.js

sigvef:
image


cristeahub:
Nice! Any way we could get some more descriptive text on the boxes? I do get the MB means allocated memory, but had to check it up and that MS means ms to render the frame, but still.

sigvef:
> Any way we could get some more descriptive text on the boxes?

Not out-of-the-box, at least. Stats.js on github: https://github.com/mrdoob/stats.js/

PR #317, opened 2017-02-05
Graph editor

sigvef:
I ripped out the old SVG graph stuff and replaced it with something more extensible. The idea is that in the future, we can use this to fully manipulate graph.json -- no need to edit it in text ever again!


aleksanb:
Travs is failing.

sigvef:
fixed linting

PR #327, opened 2017-03-15
Swap j and k keyboard shortcuts

sigvef:
Turns out this is less confusing.


stianjensen:
I agree! And the earlier we switch it the better \o/

PR #328, opened 2017-03-15
Graph improvements

aleksanb:
image

This makes the root view a bit more useful, showing connection names at the new default zoom level.

### DIFF
--- a/nin/frontend/app/scripts/directives/editor/GraphEditorInputOutput.js
+++ b/nin/frontend/app/scripts/directives/editor/GraphEditorInputOutput.js
@@ -77,9 +77,9 @@ class GraphEditorInputOutput extends React.Component {
         strokeWidth: 2,
         onClick: event => {this.props.editor.startOrCompleteConnection(this.props.item);},
       }),
-      this.props.scale > 1.5 ? e('text', {
+      this.props.scale >= 1.5 ? e('text', {
         x: 0,
-        y: 20,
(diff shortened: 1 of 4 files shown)


nin/frontend/app/scripts/directives/editor/GraphEditorInputOutput.js
```diff
@@ -77,9 +77,9 @@ class GraphEditorInputOutput extends React.Component {
         strokeWidth: 2,
         onClick: event => {this.props.editor.startOrCompleteConnection(this.props.item);},
       }),
-      this.props.scale > 1.5 ? e('text', {
+      this.props.scale >= 1.5 ? e('text', {
         x: 0,
-        y: 20,
+        y: 10 * this.props.scale,
```
sigvef: Why this change? Compare before:

image

after:

image

aleksanb: The offset is now always constant compared to the size of the node. This makes it look better at the lower zoom levels.
Font-scaling could be introduced for those who want to look really close to the nodes.

aleksanb: Right now no one uses the lower zoom levels though.
If the entire system is up for a rewrite after revision anyways, I'd rather improve it for (what i think is) the common use case: the default zoom level.

PR #333, opened 2017-03-19
Add shadernode

aleksanb:
It lives!

image

Now you can generate a ShaderNode and it will give you a Node and a Shader and it works!
The GUI will need reloading (perhaps), and you'll have to link the node up to the root node yourself.


cristeahub:
Just so I'm not mistaken: the shadernode is a node that generates a plane and passes that on, right?

stianjensen:
It renders a shader on a plane and outputs that as a texture equal to the output of most other nodes.

stianjensen:
And you can override it to make it take other nodes as input so it functions as a post processing shader.

cristeahub:
Ok, our old shaderlayers did a shaderpass instead of a renderpass as rendering technique. Haven't seen if there's any differences we should be aware of there, but I will in the next couple of days.

sigvef:
You can also use it for shader-only effects, like a raymarched scene \o/

aleksanb:
Jup!

aleksanb:
Ooh i actually have some raymarching code lyinig around. In good style i should incorporate it into our demo!

aleksanb:
https://github.com/aleksanb/raymarching/

cristeahub:
> You can also use it for shader-only effects, like a raymarched scene \o/

This shadernode can be used for that, or do we need to use a shaderpass instead?

stianjensen:
> This shadernode can be used for that, or do we need to use a shaderpass instead?

We can use this node for that.

PR #340, opened 2017-03-21
Shadernode fixes

sigvef:


cristeahub:
Why the change to the shadermaterial?

sigvef:
Sometimes we need more than just uniforms, `vertexShader`, `fragmentShader`. As an example, THREE.ConvolutionShader also uses `defines`.

sigvef:
Check out the `parameters` parameter of ShaderMaterial here for a more complete list: https://threejs.org/docs/api/materials/ShaderMaterial.html

PR #341, opened 2017-03-21
Improve graph flow rendering algorithm

sigvef:
Before (circled problem areas):
image

After:
image

### DIFF
--- a/nin/frontend/app/scripts/directives/editor/GraphEditor.js
+++ b/nin/frontend/app/scripts/directives/editor/GraphEditor.js
@@ -33,7 +33,15 @@ class GraphEditor extends React.Component {
   generateDepths() {
     let deepestLevel = 0;
     let depths = {};
-    let seen = new Set();
+
+    depths.root = {
+      x: 0,
+      y: 0,
+      id: 'root',
+      counter: 0,
+    };
+
+    let counter = 1;
 
     function recurse(level, currentNodes) {
       let nextLevel = [];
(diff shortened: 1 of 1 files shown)


nin/frontend/app/scripts/directives/editor/GraphEditor.js
```diff
@@ -35,21 +35,35 @@ class GraphEditor extends React.Component {
...
+          if(source) {
+            depths[source.node.id] = {
+              x: level + 1,
+              y: 0,
+              id: source.node.id,
+              counter
+            };
+            counter++;
+          }
+          if (source) {
```
aleksanb: We can combine these two if's'.

nin/frontend/app/scripts/directives/editor/GraphEditor.js
```diff
@@ -35,21 +35,35 @@ class GraphEditor extends React.Component {
...
             seen.add(source.node.id);
```
aleksanb: A thing i realize now is that we could drop the seen set entirely in favour of using the depths object (the keys would be the same as the contents of the set).

nin/frontend/app/scripts/directives/editor/GraphEditor.js
```diff
@@ -63,6 +77,22 @@ class GraphEditor extends React.Component {
...
+    const positions = [];
+    for(let i = 0; i <= deepestLevel; i++) {
+      positions[i] = [];
+    }
+    for(let nodeId in depths) {
+      positions[depths[nodeId].x].push(depths[nodeId]);
+    }
+
+    for(let x in positions) {
+      console.log(x, positions[x].length);
```
aleksanb: Logging can be removed.

sigvef:
updated

PR #361, opened 2017-04-03
Update fft analysis to give frequency float data

stianjensen:


nin/dasBoot/music.js
```diff
@@ -15,8 +15,11 @@ function loadMusic() {
   _gainNode.connect(webAudioContext.destination);
   var _analyserNode = webAudioContext.createAnalyser();
   _analyserNode.fftSize = 2048;
+  _analyserNode.minDecibels = -90;
```
sigvef: What is the purpose of setting minDecibels to -90 instead of the default, which is -100?

nin/dasBoot/music.js
```diff
@@ -15,8 +15,11 @@ function loadMusic() {
...
+  _analyserNode.maxDecibels = -10;
```
sigvef: Likewise for -10 vs -30 (default)?

nin/dasBoot/music.js
stianjensen: Cargo culting.

nin/dasBoot/music.js
stianjensen: Lemme take another look and consider just removing them.

stianjensen:
Ping. Removed overwriting of default values. Doesn't make much of a difference for now it seems, so I'll spend time looking into if they need tweaking later instead.

PR #365, opened 2017-04-08
Fix BEATBEAN ordering when compiling

sigvef:
This makes BEATBEAN properly available in Node constructors in compiled
demos.


stianjensen:
Does this mean we can also stop defining BEAT and BEAN in compile.js?

sigvef:
probably?

PR #367, opened 2017-04-09
Fix system specific path breaking Windows reloading on save.

Raane:
This will ensure that the path of reloading files are not Windows formated.

### DIFF
--- a/nin/backend/watch.js
+++ b/nin/backend/watch.js
@@ -44,7 +44,7 @@ function watch(projectPath, cb) {
                   chalk.magenta(path));
     }
 
-    cb(event, {path: path});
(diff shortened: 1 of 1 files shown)


nin/backend/watch.js
```diff
@@ -44,7 +44,7 @@ function watch(projectPath, cb) {
                   chalk.magenta(path));
     }
 
-    cb(event, {path: path});
+    cb(event, {path: path.replace(/\\/g,"/")});
```
stianjensen: Please change double quote to single quote. Travis complains :)

sigvef: Since we're changing a linting thing, we might as well add a space after the comma too :)

aleksanb:
Superseded by #400

PR #377, opened 2017-04-15
Use path.sep for cross-platform path splitting

stianjensen:


iver56:
I'm not able to test this atm. If it works for you, go ahead and merge

aleksanb:
I think you want to use path.parse(thepath).base instead, as documented here: https://nodejs.org/api/path.html.

stianjensen:
> I think you want to use path.parse(thepath).base instead, as documented here: 

That won't work, since the file that is being changed has a path like this:
`src/shaders/yolo/fragment.glsl`

`base` will then be `fragment.glsl`, and `dir` will be `src/shaders/yolo`.

aleksanb:
Right now we're only using the second segment from the shader path (pathParts[2]), wouldn't those two be equivalent then

stianjensen:
`'src/shaders/yolo/fragment.glsl'.split('/')` equals `['src', 'shaders', 'yolo', 'fragment.glsl']`, so `[2]` will then be `yolo`, while `base` will be `fragment.glsl`.

aleksanb:
Hm. Doesn't seem like node's path has any built-in splitting thing (and just recommends you to split yourself).

stianjensen:
> Hm. Doesn't seem like node's path has any built-in splitting thing (and just recommends you to split yourself).

Correct:
https://github.com/nodejs/node/issues/5523

PR #380, opened 2017-04-15
Always compile .png.html and .html

aleksanb:
Now we always get both versions, as there's very little overhead in writing the .html version once we have the intermediate step after the closure compiler has ran.

I've spent quite some time building both versions, and now we always get both!


stianjensen:
It's still called `.png.html` and not `.html.png` 🤓

PR #381, opened 2017-04-15
Fix sorting function for project files

stianjensen:
This should resolve our spurious issues with loading nin projects!


stianjensen:
ping, removed console.logs

PR #382, opened 2017-04-16
Add option for only closure compiling whitespace

stianjensen:
On my machine this reduces compile time from 2 minutes to 1 minute.
Meanwhile, the filesize increased (in the one example I tested with) from
5.2 MB to 5.3 MB.

This will be nice for incremental testing of compiles etc.

### DIFF
--- a/nin/backend/compile.js
+++ b/nin/backend/compile.js
@@ -179,6 +179,7 @@ const compile = function(projectPath, options) {
               path
             }));
           const out = closureCompiler.compile({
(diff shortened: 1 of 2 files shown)


stianjensen:
Actually, there might be something broken here. I'll investigate and come back.

stianjensen:
Yeah, turns out this doesn't actually work. When using whitespace only with closure compiler, es6 transpiling fails, and it doesn't support outputting ES6 code yet.

PR #386, opened 2017-04-16
Faster compile

stianjensen:


nin/backend/nin
```diff
@@ -26,9 +26,8 @@ program
   .command('compile')
   .description('Compile the nin project')
   .option('--no-optimize-png-assets', 'Do not use optipng to optimize assets')
-  .action(options => compile.compile(utils.findProjectRootOrExit(process.cwd()), {
-    optimizePngAssets: options.optimizePngAssets
-  }));
+  .option('--fast', 'Bypass Google Closure Compiler (faster)')
```
sigvef: Perhaps `--no-closure-compiler` instead? Closure compiler is not the only one that makes things fast/slow.

stianjensen:
ping

stianjensen:
ping

PR #387, opened 2017-04-16
Node 7.9

aleksanb:


stianjensen:
You can specify required node version in package.json as well! Then npm will hopefully give some sort of helpful warning when installing:
https://docs.npmjs.com/files/package.json#engines

stianjensen:
conflicts

PR #388, opened 2017-04-17
Fix shadergen import

aleksanb:


stianjensen:
s/shadergen/dasbootgen/ ¯\_(ツ)_/¯

sigvef:
Self merge?

stianjensen:
Oh also that. Need to detox from the demo crunching!

PR #392, opened 2017-04-18
Update README

cristeahub:
Not sure about the wording, come with suggestions on how to possibly improve it.


stianjensen:
How about something like:

```

cristeahub:
i like it, ping

PR #397, opened 2017-04-22
Add better font support

sigvef:
Fonts can now be added to the res/ folder, and they will be
automagically be made available as a `@font-face` with the same name as
the font file (without suffix). As an example: `res/myfont.otf` will be
available for a 2D canvas as `ctx.font = '10pt myfont';`. Only .otf is
supported for now.

Before this commit, the typical way of adding fonts was to manually
convert otf to base64 offline, load that with Loader.loadAjax, and take
care to create a style tag with the font definition somewhere in the
demo.

Some font loaded in some demo:
image

Fonts being skipped when assimilating res:
image

### DIFF
--- a/nin/backend/compile.js
+++ b/nin/backend/compile.js
@@ -10,6 +10,7 @@ const shaderGen = require('./shadergen').shaderGen;
 const stream = require('stream');
 const utils = require('./utils');
 const dasbootGen = require('./dasbootgen');
+const fontGen = require('./fontgen');
 const walk = require('walk');
 
 
@@ -44,7 +45,14 @@ function res(projectPath, options, callback) {
     root += '/';
     root = root.replace(/\/\//g, '/');
 
-    const file = await fs.readFile(root + stat.name);
+    const filename = root + stat.name;
+    const file = await fs.readFile(filename);
+    if(p.extname(filename) == '.otf') {
+      process.stdout.write('- Skipping ' + chalk.grey(root.slice(projectPath.length + 1)) + chalk.magenta(stat.name) + chalk.grey(', (it is a font)'));
+      renderOK();
+      next();
+      return;
+    }
     process.stdout.write('- Assimilating ' + chalk.grey(root.slice(projectPath.length + 1)) + chalk.magenta(stat.name));
     function pushFinishedFile(file) {
       files.push('FILES[\'' + root.slice(projectPath.length + 1) + stat.name + '\']=\'' +
@@ -157,7 +165,10 @@ const compile = function(projectPath, options) {
 
     projectSettings.generate(projectPath);
     shaderGen(projectPath, async function() {
-      await dasbootGen(projectPath);
+      await Promise.all([
+        fontGen(projectPath),
+        dasbootGen(projectPath),
+      ]);
 
       process.stdout.write(chalk.yellow('\nRunning closure compiler'));
       const globPaths = [
--- a/nin/backend/fontgen.js
+++ b/nin/backend/fontgen.js
@@ -1,0 +1,43 @@
+const fs = require('fs-promise');
+const glob = require('glob');
+const path = require('path');
+
+
+async function fontGen(pathPrefix) {
+
+  let resolver;
+  let promise = new Promise(resolve => {
+    resolver = resolve; 
+  });
+
+  glob(path.join(pathPrefix, 'res/**/*.otf'), {}, async (error, filenames) => {
+    let fonts = {};
+    for(const filename of filenames) {
+      const file = fs.readFileSync(filename);
+      const name = path.basename(filename, '.otf');
+      fonts[name] = file.toString('base64');
+    }
+
+    const fontjs = `(() => {
+      const fonts = ${JSON.stringify(fonts)};
+      for(const name in fonts) {
+        const id = name + '-font';
+        const font = fonts[name];
+        const s = document.createElement('style');
+        s.setAttribute('id', id);
+        s.innerHTML = [
+          '@font-face {',
+          'font-family: "' + name + '";',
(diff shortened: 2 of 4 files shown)


sigvef:
Ignore whitespace-only lines for easier diff reading: https://github.com/ninjadev/nin/pull/397/files?w=1

stianjensen:
Needs rebase

nin/backend/compile.js
```diff
@@ -46,7 +47,14 @@ function res(projectPath, options, callback) {
     root += '/';
     root = root.replace(/\/\//g, '/');
 
-    const file = fs.readFileSync(root + stat.name);
+    const filename = root + stat.name;
+    const file = fs.readFileSync(filename);
+    if(filename.slice(-4) == '.otf') {
```
aleksanb: You can use path.extname instead here.

nin/backend/fontgen.js
```diff
@@ -0,0 +1,38 @@
+const fs = require('fs');
+const glob = require('glob');
+const path = require('path');
+
+
+let fontGen = function(pathPrefix, cb) {
+
+  glob(path.join(pathPrefix, 'res/**/*.otf'), {}, (error, filenames) => {
```
aleksanb: Why /**/*.otf and not just /*.otf ?

nin/backend/fontgen.js
```diff
@@ -0,0 +1,38 @@
+const fs = require('fs');
+const glob = require('glob');
+const path = require('path');
+
+
+let fontGen = function(pathPrefix, cb) {
```
aleksanb: I reccomend you switch fontGen to be async, so that we get away with one less indentation level in compile.js (soon our callbacks will be larger than the pyramids of Giza!).
you can switch fs.writeFileSync with await fs.outputFile from https://github.com/jprichardson/node-fs-extra/blob/master/docs/outputFile.md.

nin/backend/fontgen.js
```diff
@@ -0,0 +1,38 @@
...
+        s.innerHTML = [
+          '@font-face {',
+          'font-family: "' + name + '";',
+          'src: url(data:application/x-font-opentype;charset=utf-8;base64,' + font + ') format("opentype");',
+          '}'
+        ].join('\\n');
+        document.body.appendChild(s);
+      }
+    })();`;
+    fs.writeFileSync(path.join(pathPrefix, 'gen/fonts.js'), fontjs);
```
aleksanb: You should skip the '/' and join pathPrefix, gen, fonts for cleanliness.

nin/backend/serve.js
```diff
@@ -18,6 +19,7 @@ const serve = function(
   const genPath = p.join(projectPath, 'gen');
   mkdirp.sync(genPath);
 
+  fontGen(projectPath, () => {});
```
aleksanb: can be replaced with await fontGen after change.

sigvef:
> Needs rebase

rebased

sigvef:
> What happens if you add a font to a running nin project?
> Should probably add something to watcher.js to handle live font updating.

We don't support live reloading of most resources, so I think not supporting it for fonts too is ok.

nin/backend/compile.js
sigvef: done

nin/backend/fontgen.js
sigvef: yep, now that the other functions are async it makes sense.

nin/backend/fontgen.js
sigvef: double star is glob for "recursive".

nin/backend/fontgen.js
sigvef: done

nin/backend/serve.js
sigvef: yep

nin/backend/fontgen.js
```diff
@@ -0,0 +1,44 @@
+const fs = require('fs-promise');
+const glob = require('glob');
+const path = require('path');
+
+
+async function fontGen(pathPrefix) {
+
+  let resolver;
+  let promise = new Promise(resolve => {
```
aleksanb: Is this code equivalent to wrapping the glob inside the `new Promise(resolve => {
}` code? Can there be a race condition  now that the scheduler runs the glob callback before the callback given to the promise, so resolver is undefined?

sigvef: I think it is equivalent to wrapping the glob inside the promise callback. I split it like this to reduce indentation.

aleksanb: I agree it's nice to reduce rightwards drift, although i find having everything inside the promise to be more readable.

Piece by piece fleshing out the Ninjadev style guide 💃 @stianjensen ?

stianjensen: I think this is a nice tradeoff to avoid multi-level nesting. One alternative is switching out `glob` for `globby`, which gives us a nice promise based api instead of the callback to the glob function.

aleksanb: https://github.com/sindresorhus/globby/blob/master/index.js#L67

Probably easier just to do it ourselves.

nin/backend/fontgen.js
```diff
@@ -0,0 +1,44 @@
...
+      const name = path.basename(filename, '.otf');
+      fonts[name] = file.toString('base64');
+    }
+
+    const fontjs = `(() => {
+      const fonts = ${JSON.stringify(fonts)};
+      for(const name in fonts) {
+        const id = name + '-font';
+        const font = fonts[name];
+        console.log(font);
```
stianjensen: Remove logging?

sigvef: removed

sigvef:
ping

PR #398, opened 2017-04-22
Add shader optimization step to build pipeline

aleksanb:
Currently only strips whitespace, and only for nin compile.
Even with just whitespace strip, we get a savings of 340kb -> 31kb = 309kb.

The pluggable GLSL minifier is here: https://github.com/stackgl/glsl-min-stream, but the variable renaming is massively broken and i'm not up for fixing it yet.


stianjensen:
Needs a lil' rebase

nin/backend/shadergen.js
```diff
@@ -1,67 +1,74 @@
...
-
-  traversePath(pathPrefix + '/src/shaders/', function() {
-    traversePath(__dirname + '/../dasBoot/shaders/', function() {
-      fs.writeFileSync(pathPrefix + '/gen/shaders.js', out);
-      cb(out);
-    });
+  traversePath(p.join(pathPrefix, 'src', 'shaders'), () => {
+    fs.writeFileSync(p.join(pathPrefix, 'gen', 'shaders.js'),
+      'SHADERS='+JSON.stringify(out));
+    cb();
```
stianjensen: Missing `out` in callback here?

PR #399, opened 2017-04-22
Switch to sane fs

aleksanb:


nin/backend/compile.js
```diff
@@ -155,47 +146,46 @@ const compile = function(projectPath, options) {
...
-            renderWarn();
-            out.warnings.map(console.error);
-          } else {
-            renderOK();
-          }
-
-          collect(out.compiledCode);
-        });
+    await fs.remove(genPath);
+    await fs.outputFile(p.join(genPath, 'files.js'), data);
```
stianjensen: outputFile doesn't require creating a new Buffer, like we did with writeFile?

PR #400, opened 2017-04-22
Fix windows / unix path issues in single place

aleksanb:


stianjensen:
LGTM.

@iver56 @Raane confirm?

stianjensen:
Never mind, it probably Just™ Works™.

PR #404, opened 2017-04-23
Stop demo properly when it reaches end

stianjensen:
This fixes two issues:

1. When the demo reaches the end during playback, it sometimes ended on a decimal-numbered frame. Frame numbers will with this fix always be integers.
2. When rendering a demo to images, the rendering never stopped, even after the end of the song. This is now resolved.

### DIFF
--- a/nin/dasBoot/bootstrap.js
+++ b/nin/dasBoot/bootstrap.js
@@ -96,7 +96,7 @@ window['bootstrap'] = function(options) {
     var time = (frame / 60) * 1000;
     if (time > demo.music.getDuration() * 1000) {
       time = demo.music.getDuration() * 1000;
-      frame = (time / 1000) * 60;
+      frame = (time / 1000) * 60 | 0;
     }
     else if (frame < 0) {
       frame = 0;
--- a/nin/frontend/app/scripts/utils/render.js
+++ b/nin/frontend/app/scripts/utils/render.js
@@ -1,7 +1,6 @@
 const demo = require('../demo');
 const commands = require('../commands');
 
-let currentFrame;
 let currentlyRendering = false;
 let currentTimeout;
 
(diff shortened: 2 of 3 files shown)


nin/dasBoot/bootstrap.js
```diff
@@ -96,7 +96,7 @@ window['bootstrap'] = function(options) {
     var time = (frame / 60) * 1000;
     if (time > demo.music.getDuration() * 1000) {
       time = demo.music.getDuration() * 1000;
-      frame = (time / 1000) * 60;
+      frame = (time / 1000) * 60 | 0;
```
sigvef: There should probably be a `+ 1` here to make sure we don't drop up to 16.6666..ms of music?

nin/frontend/app/scripts/utils/render.js
```diff
@@ -19,8 +18,13 @@ commands.on('stopRendering', () => {
 
 function render(i) {
   i = i || 0;
-  currentFrame = i;
   demo.jumpToFrame(i);
+  const currentFrame = demo.getCurrentFrame();
+  if (currentFrame < i) {
```
sigvef: perhaps this should be `currentFrame <= i` so that we can render the final frame of the demo as well?

nin/frontend/app/styles/common.less
```diff
@@ -76,6 +76,7 @@ main-window, main-window > div {
   height: 50px;
   position: absolute;
   bottom: 0;
+  overflow: hidden;
```
sigvef: What was this needed for?

stianjensen: To avoid a crazy scrollbar dance at the end of the demo

nin/frontend/app/scripts/utils/render.js
stianjensen: The normal case would be that `currentFrame == i`, no? So that would stop rendering at the first frame.

sigvef: Yes, I misread the if body.

PR #407, opened 2017-04-23
Music bar styles

sigvef:
This adds bpm-based lines to the music view, and some additional music
variable info panels.

image

### DIFF
--- a/nin/backend/projectSettings.js
+++ b/nin/backend/projectSettings.js
@@ -9,7 +9,8 @@ const defaultSettings = {
   music: {
     path: 'res/music.mp3',
     bpm: 125,
-    subdivision: 4
+    subdivision: 4,
+    beatOffset: 0,
   },
   googleAnalyticsID: 'Your GA ID'
 };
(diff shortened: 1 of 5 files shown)


aleksanb:
image

The bars seem to be all over the place.
In addition, when you set a loop, the green bars are not placed at the same position as the other bars.
Presumably they should use the same form of bar computation.

stianjensen:
> In addition, when you set a loop, the green bars are not placed at the same position as the other bars.
> Presumably they should use the same form of bar computation.

The bar lines are every fourth beat, while the loop point can snap to every beat, maybe?

sigvef:
> The bars seem to be all over the place.

What do you mean?

sigvef:
> In addition, when you set a loop, the green bars are not placed at the same position as the other bars.

The green bars are snapped to quavers (i.e., one fourth of a bar), and the lines are snapped to bars.

sigvef:
> There seems to be a ton of logging going on:

I don't think so? Are you on the right branch?

sigvef:
I'm changing the rendering strategy, so closing this for now.

PR #409, opened 2017-04-24
Switch font loader to use woff2 instead of otf

stianjensen:
otf files have no compression while woff and woff2 provide much smaller
font files. The one example I tested on (the font vcr from `re`) was
half the filesize of the corresponding otf file, and reduced the size of
the .png-compiled demo by almost 2 kB (that's almost half a 4k demo!).

Woff2 also has basically as good browser support as otf, only lacking
internet explorer (but edge works!), and Safari for those not running
the latest macOS/iOS. Before we release another prod, I don't think that
will be an issue.

### DIFF
--- a/nin/backend/compile.js
+++ b/nin/backend/compile.js
@@ -47,7 +47,7 @@ function res(projectPath, options, callback) {
 
     const filename = root + stat.name;
     const file = await fs.readFile(filename);
-    if(p.extname(filename) == '.otf') {
+    if(p.extname(filename) == '.woff2') {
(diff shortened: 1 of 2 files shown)


sigvef:
Did you see this working? It doesn't work with a .woff2 font that I converted from a working .otf using fontforge, at least.

stianjensen:
Huh, yeah I tested it with the vcr font we have. I can a push a branch with that and we can see if there's an issue with the font or something like the browser.

stianjensen:
The branch I tested with: https://github.com/ninjadev/re/tree/woff2

sigvef:
I tried with a different woff2, and it works. I probably had a bad font file the last time I tested.

PR #411, opened 2017-04-24
Fix loading of muted state from local storage

stianjensen:
Turns out we have kept around a volume parameter for a while without
ever having a possibility to set the volume to anything other than 1 or
0. I take this as a hint that this functionality is not needed and
simplified the muting a bit in the process.

### DIFF
--- a/nin/frontend/app/scripts/components/main.js
+++ b/nin/frontend/app/scripts/components/main.js
@@ -27,13 +27,10 @@ class Main extends React.Component {
       fullscreen: false,
       showFramingOverlay: false,
       mute: localStorage.getItem('nin-mute') ? true : false,
-      volume: 1,
       globalJSErrors: {},
     };
 
-    if (localStorage.hasOwnProperty('nin-volume')) {
-      this.state.volume = +localStorage.getItem('nin-volume');
-    }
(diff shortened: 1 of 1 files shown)


nin/frontend/app/scripts/components/main.js
```diff
@@ -27,13 +27,10 @@ class Main extends React.Component {
...
       showFramingOverlay: false,
       mute: localStorage.getItem('nin-mute') ? true : false,
-      volume: 1,
       globalJSErrors: {},
     };
 
-    if (localStorage.hasOwnProperty('nin-volume')) {
-      this.state.volume = +localStorage.getItem('nin-volume');
-    }
+    demo.music.setVolume(1 - this.state.mute);
```
aleksanb: Here you use 1 - mute, but further down you use negation.

stianjensen: I use negation to toggle the boolean mute flag, while setVolume takes a number between 0 and 1. I suppose use negation will technically work here too, since true is 1 and false is 0, but I felt it was better to actually pass a number instead.

sigvef:
fwiw this was written like this so that mute wouldn't clobber your volume setting if you have sneakily set your own using the web inspector, iirc.

PR #413, opened 2017-04-24
Migrate dasboot to require

aleksanb:
This is step 1 / N of transforming our code base so we can use the google closure compiler advanced settings.


nin/frontend/package.json
```diff
@@ -25,8 +25,7 @@
     "less": "^2.7.2",
     "less-loader": "^3.0.0",
     "ng-annotate-loader": "^0.2.0",
-    "style-loader": "^0.14.0",
-    "webpack": "^2.2.1"
```
stianjensen: I think removing webpack here introduces more issues than it solves

nin/dasBoot/bootstrap.js
```diff
@@ -1,3 +1,53 @@
+window.THREE = require('./lib/00_three');
```
sigvef: Can rename to not have numeric prefixes now, perhaps

aleksanb: This comes in a later commit.

aleksanb:
There is a performance hit on nin run of < 1 second.
Should be fine given that you're not nin running all the time.

aleksanb:
Ping.

stianjensen:
LGTM. @sigvef ?

sigvef:
> There is a performance hit on nin run of < 1 second.
> Should be fine given that you're not nin running all the time.

I was thinking more of demo runtime performance. Does this impact running demos (especially when compiled?).

sigvef:
I get this when running on `re`:

image

image

aleksanb:
Don't think it should.
If anything, it would only impact load times,
as there's a call to __webpack_require to load a module now, as opposed to it being in global scope. But all the important variables are still bound to window, so playback performance should be the same.

There is a slight size increase in the demo (5kb?) but with future advanced optimizations the total size should go way down.

PR #417, opened 2017-04-25
Add truck-nin as separate command

aleksanb:
image

This feature has been severely lacking from our lives.


cristeahub:
But are we ready for such a powerful feature?

PR #422, opened 2017-04-30
Add scrollwheel scrubbing and volume control

joms:
Add
- Scrollwheel scrubbing on waveform layer
- Range control for controlling audio volume
- Function for getting the current volume
- LESS Class for rotating range selector

Change
- Make scrollwheel event fire only on relevant elements
  - Scrollwheel for grapheditor now fires only on grapheditor
- Make `.bottom` footer fill the whole area, whilst content inside can be adjusted


stianjensen:
Thanks for this!

Only thing I noticed is that when lowering the volume and muting/unmuting, the volume is incorrect after unmuting. This should be a simple fix, just getting the actual volume here instead of setting it to 1: https://github.com/ninjadev/nin/blob/master/nin/frontend/app/scripts/components/main.js#L78

joms:
Thanks for highlighting this issue. Added a fix for this

Add
- Add a `state.volume` parameter

Change
- Make `toggleMusic` event save current volume to `state.volume`

stianjensen:
> Thanks for highlighting this issue. Added a fix for this

Coolio!

Sorry for my slow response btw, the weather has just been too good this week 😅 

Anyway, I just noticed a small issue when testing the latest change now, which confuses me because I'm pretty sure I didn't see it when I tested a few days ago. The waveform fills the whole screen now, even behind the volume control. Is it like that on your end as well?

Screenshot:
screenshot 2017-05-06 17 29 39

stianjensen:
> Anyway, I just noticed a small issue when testing the latest change now, which confuses me because I'm pretty sure I didn't see it when I tested a few days ago. The waveform fills the whole screen now, even behind the volume control. Is it like that on your end as well?

Turns out this error only shows up on retina screens, hence why I didn't see it the other day. That problem is not directly related to this, so I'll fix that separately.

Thanks for making this feature! 🎉

joms:
What's this outside you talk of?!

Awesome that you figured it out! Kind of weird that it happens though. I think maybe the layout markup should be rewritten to something more like a grid to support changes and display sizes better. I guess it'd be a pretty big job though.   
Also, the way I made the waveform narrower might not have been the best way.

stianjensen:
> Also, the way I made the waveform narrower might not have been the best way.

I made a fix for it now: https://github.com/ninjadev/nin/pull/423

Still not very pretty with hard-coding the same 50px in multiple files, but as you say most of the components are just fixed-positioned around currently.

PR #424, opened 2017-05-07
Replace zlib deflate with zopfli (free filesize reduction!)

stianjensen:
Zopfli is a zlib-compatible compression algorithm, which means it can be
used as a drop-in replacement inside PNG, while giving generally better
compression.

Testing on ninjadev/re shows a 60kB filesize decrease when using zopfli
instead of zlib.

The node-zopfli package has bindings to a zopfli c-implementation, so hopefully it's not much hassle to install on windows. The documentation suggests it should work fine, though.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,3 +16,7 @@ addons:
       - ubuntu-toolchain-r-test
     packages:
      - g++-4.8
+before_install:
+  - export CXX=g++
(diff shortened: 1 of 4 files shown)


stianjensen:
I'll investigate replace the node-zopfli dependency with an emscripten compiled version to make it less hassle to install.

PR #427, opened 2017-06-17
Re-render current frame after shader or node change

lionleaf:
Recompiling shaders before this fix would result in a white frame if
paused.


lionleaf:
ping

stianjensen:
I can't really see a difference in behavior with/without this fix. Do you have a suggestion for how to test it?

sigvef:
needs rebase

sigvef (changes requested):
Rejecting for colorizer (because needs rebase).

sigvef (approved):

PR #440, opened 2017-07-17
Start gradual typing of frontend

aleksanb:
This commit introduces Typescript to our Webpack 2.0 flow.
Advantages include help from the compiler in weeding out logic errors in
components/main.js (i've simplified the websocket event type already),
and better compile time verification of our code.

I've not added ts to the backend (yet), as i couldn't figure out a
satisfactory build system just yet (just putting the compiled .js in
backend/dist could probably work).

I've yet to find a proper way of sharing files across the two projects,
as i initallly wanted to share the WSEvent type, so that i'd be
guaranteed the data sent over the socket matched the structure of the
one received.

### DIFF
--- a/nin/frontend/app/scripts/components/main.tsx
+++ b/nin/frontend/app/scripts/components/main.tsx
@@ -1,5 +1,3 @@
-const React = require('react');
-const SocketController = require('../socket');
 const Bottom = require('./bottom');
 const MenuBar = require('./menubar');
 const GraphEditor = require('./editor/GraphEditor');
@@ -8,7 +6,17 @@ const commands = require('../commands');
 const demo = require('../demo');
 require('../utils/render');
 
-class Main extends React.Component {
+import * as React from "react";
+import {SocketController, WebSocketEvent} from '../socket';
+
+declare const Loader: any;
+declare const NIN: any;
+
+export default class Main extends React.Component<any, any> {
+  themes: string[];
+  fileCache: Object;
+  startupSound: any;
+
   constructor(props) {
     super(props);
 
@@ -73,7 +81,7 @@ class Main extends React.Component {
     commands.on('toggleMusic', () => {
       if (!this.state.mute) {
         this.setState({volume: demo.music.getVolume()});
-        localStorage.setItem('nin-mute', 1);
+        localStorage.setItem('nin-mute', '1');
         demo.music.setVolume(0);
       } else {
         localStorage.removeItem('nin-mute');
@@ -99,7 +107,7 @@ class Main extends React.Component {
     }
 
     var layerShaderDependencies = {};
-    socketController.on('add change', event => {
+    socketController.on('add change', (event: WebSocketEvent) => {
       try {
         switch (event.type) {
         case 'graph':
@@ -140,7 +148,7 @@ class Main extends React.Component {
         case 'camera':
           for (let key in demo.nm.nodes) {
             if (demo.nm.nodes[key] instanceof NIN.THREENode) {
-              if (demo.nm.nodes[key].options.camera == event.path) {
+              if (demo.nm.nodes[key].options.camera == event.name) {
                 demo.nm.nodes[key].initializeCamera(event.content);
               }
             }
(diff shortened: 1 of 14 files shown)


nin/frontend/app/scripts/components/main.tsx
```diff
@@ -183,15 +191,14 @@ class Main extends React.Component {
...
+        delete globalJSErrors[event.name];
         this.setState({globalJSErrors});
       } catch (e) {
         e.context = "WS load of " + event.name + " failed";
-        e.type = event.type;
         e.name = event.name;
 
         const globalJSErrors = Object.assign({}, this.state.globalJSErrors);
-        globalJSErrors[event.type] = e;
+        globalJSErrors[event.name] = e;
```
stianjensen: changing from type to kind vs name here. What impact does this have on the behavior?

sigvef:
s/initallly/initially/ (commit message)

nin/frontend/app/scripts/components/main.tsx
sigvef: Why are we changing it? And can't we just use the type system directly?

nin/frontend/app/scripts/socket.ts
```diff
@@ -1,6 +1,15 @@
 const SockJS = require('sockjs-client');
 
-class SocketController {
+export interface WSEvent {
```
sigvef: Perhaps we can call this WebSocketEvent for increased clarity?

nin/frontend/app/scripts/components/main.tsx
aleksanb: I switched back to using type as the name.

nin/frontend/app/scripts/socket.ts
aleksanb: Renamed.

aleksanb:
Regarding using a union type (as currently used) versus inheritance to discriminate the websocket types,
union types with an explicit type attribute has the advantage of simplifying deserialization of events coming over the wire.
When events are sent from the backend, we either need to specify which kind of type the event is, or some clever serialization scheme for something like `class GraphEvent extends WebSocketEvent` so that we can deserialize again on the frontend.

Continuing to use the `type` attribute feels like less friction, and a better fit for the events we send.
In addition, using a sum type allows us to have different fields for the events sent in the future, as even though the field names for all variants are the same right now, they might not be so in the future.

sigvef:
> In addition, using a sum type allows us to have different fields for the events sent in the future, as even though the field names for all variants are the same right now, they might not be so in the future.

This is true for inheritance too, so there is no difference there.

> When events are sent from the backend, we either need to specify which kind of type the event is, or some clever serialization scheme for something like class GraphEvent extends WebSocketEvent so that we can deserialize again on the frontend.

I get the feeling that somewhere down the line we're probably going to want to come up with serialization anyway as we add more websocket events, but that's more of a future issue.

In the end, the pros and cons for the two approaches aren't really that different, and we can always change it later again if need be! Now to actually test this thing and see if it works on my machine :D

sigvef:
when running make:

image

sigvef:
node 7.9.0, for reference

aleksanb:
Should work now. I've had to disable uglify as it can't handle any newer than ES5 (so no native async/await in the frontend then).
In addition, there were so many indentation errors which showed up now for some reason, i've changed indentation to warn from err in eslintrc.

stianjensen:
Is eslint running before or after typescript has altered the source?

stianjensen:
You disabled uglify by not setting NODE_ENV to production? Make note that this also disables the production version of react: http://dev.topheman.com/make-your-react-production-minified-version-with-webpack/

aleksanb:
I've updated the PR to make Typescript target es5.
This is required as long as we want to keep using uglify (which only handles es5 at the moment).
Uglify turns out to be really difficult to remove from webpack, and we only needed minimal tweaks to the codebase to get it to transpile to es5.

Another workaround is disabling webpack production mode, but this also disables react optimization, reducing performance again.
As soon as webpack updates to uglify 2/3 we can switch back to targeting es2017 or something like that.

PR #446, opened 2017-07-30
Auto package

sigvef:


nin/backend/compile.js
```diff
@@ -32,18 +33,44 @@ function renderError() {
...
+function slugifyTitle(title) {
+  return slugify(title).toLowerCase();
+}
+
+async function createArchive(projectPath) {
+  return new Promise(resolve => {
+    process.stdout.write(chalk.yellow('\nCreating zip archive'));
+    const metadata = utils.getProjectMetadata(projectPath);
+    const slug = slugifyTitle(metadata.projectSettings.title);
+    const outputStream = fs.createWriteStream('bin/' + slug + '.zip');
```
aleksanb: Here and other  places you create paths you should use p.join to get the correct path separators for windows and GNU/linux.

nin/backend/compile.js
```diff
@@ -32,18 +33,44 @@ function renderError() {
...
+    const archive = archiver('zip');
+    archive.pipe(outputStream);
+    let foundErrors = false;
+    archive.on('warning', e => {
```
aleksanb: Alternatively reject the promise with the error message here,
and move error handling to a single place further out in the code.

nin/backend/compile.js
```diff
@@ -83,12 +110,14 @@ async function res(projectPath) {
 }
 
 async function collect(projectPath, data) {
-  process.stdout.write(chalk.yellow('\nGenerating demo.html'));
+  const projectMetadata = utils.getProjectMetadata(projectPath);
+  const slug = slugifyTitle(projectMetadata.projectSettings.title);
+  process.stdout.write(chalk.yellow('\nGenerating ' + slug + '.html'));
```
aleksanb: No need to get the slug separately here, you can just `projectSettings.title` below the destructuring.

nin/backend/compile.js
sigvef: Either way is fine. The way it is done now is consistent with the rest of the error handling in this file, so I think it is best to leave it as-is for now.

nin/backend/compile.js
sigvef: What do you mean?

nin/backend/compile.js
sigvef: fixed

sigvef:
ping

PR #449, opened 2017-10-15
Add first steps towards an ui-editable graph

sigvef:
Now graph edges can be added and removed in a crude first-step
implementation of an editable ui. The plan is to slowly iterate towards
a slick solution over multiple pull requests.

This pull requests disables live watch reloads on res/graph.json. The
idea is that it is loaded from the backend only once, and from then on
the frontend is responsible for editing it, sending graph change events
to the backend for persistance.

### DIFF
--- a/nin/backend/socket.js
+++ b/nin/backend/socket.js
@@ -37,6 +37,20 @@ function socket(projectPath, onConnectionCallback) {
     conn.on('data', function (message) {
       let event = JSON.parse(message);
       switch (event.type) {
+        case 'graph-disconnect':
+          graph.transform(projectPath, g => {
+            const index = g.findIndex(nodeInfo => nodeInfo.id === event.data.toNode);
+            delete g[index].connected[event.data.toInput];
+          });
+          break;
+
+        case 'graph-connect':
+          graph.transform(projectPath, g => {
+            const index = g.findIndex(nodeInfo => nodeInfo.id === event.data.toNode);
+            g[index].connected[event.data.toInput] = event.data.fromNode + '.' + event.data.fromOutput;
+          });
+          break;
+
         case 'set':
           // TODO: Untested, nothing uses this yet
           graph.transform(projectPath, function(g) {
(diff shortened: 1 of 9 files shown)


sigvef:
The rendering is a little inaccurate, I'm thinking we can fix that later :3

nin/backend/socket.js
```diff
@@ -36,7 +36,22 @@ function socket(projectPath, onConnectionCallback) {
 
     conn.on('data', function (message) {
       let event = JSON.parse(message);
+      console.log('incoming!', event);
```
stianjensen: console.log

nin/frontend/app/scripts/components/editor/GraphEditor.jsx
```diff
@@ -55,6 +57,72 @@ class GraphEditor extends React.Component {
     this.inspectedItem = null;
   }
 
+  removeConnection(from, to) {
+    console.log('removing', from, to);
```
stianjensen: console.log

nin/frontend/app/scripts/components/editor/GraphEditor.jsx
```diff
@@ -55,6 +57,72 @@ class GraphEditor extends React.Component {
...
+          toNode,
+          toInput,
+        });
+        break;
+      }
+    }
+  }
+
+  connectClick(event, io) {
+    console.log(event, io);
```
stianjensen: console.log

nin/frontend/app/scripts/components/editor/GraphEditor.jsx
```diff
@@ -55,6 +57,72 @@ class GraphEditor extends React.Component {
...
+    if(!this.state.connectionStart) {
+      const connectionStart = {
+        x: io.props.node.props.x + io.props.x,
+        y: io.props.node.props.y + io.props.y,
+      };
+      console.log(connectionStart);
```
stianjensen: console.log

nin/frontend/app/scripts/components/editor/GraphEditorInputOutput.js
```diff
@@ -36,7 +36,10 @@ class GraphEditorInputOutput extends React.Component {
         style: {
           cursor: 'pointer',
         },
-        onClick: event => {this.props.editor.inspect(this);},
+        onClick: event => {
+          this.props.editor.connectClick(event, this);
+          //this.props.editor.inspect(this);
```
stianjensen: Does this mean we no longer have inspecting of nodes?

PR #452, opened 2017-10-28
Upgrade THREE from v84 to v87

sigvef:


stianjensen:
> Do you also need to bump some of the extra three files we've embedded?

Turns out these were removed previously, so not necessary.

PR #459, opened 2017-11-16
Fix bug with insertOrReplace

sigvef:
A node that was going to be replaced did not get it's inputs linked
properly. This commit fixes this issue.

### DIFF
--- a/nin/dasBoot/NodeManager.js
+++ b/nin/dasBoot/NodeManager.js
@@ -5,22 +5,29 @@ class NodeManager {
   }
 
   createNode(nodeInfo) {
-    nodeInfo.options = nodeInfo.options || {};
+    const deepClonedNodeInfo = JSON.parse(JSON.stringify(nodeInfo));
+    deepClonedNodeInfo.options = deepClonedNodeInfo.options || {};
 
-    const nodeConstructor = nodeInfo.type.slice(0, 4) === 'NIN.' ?
-      NIN[nodeInfo.type.slice(4)]
-      : window[nodeInfo.type];
+    const nodeConstructor = deepClonedNodeInfo.type.slice(0, 4) === 'NIN.' ?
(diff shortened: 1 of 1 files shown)


nin/dasBoot/NodeManager.js
```diff
@@ -5,22 +5,29 @@ class NodeManager {
   }
 
   createNode(nodeInfo) {
-    nodeInfo.options = nodeInfo.options || {};
+    const deepClonedNodeInfo = JSON.parse(JSON.stringify(nodeInfo));
```
aleksanb: Is parse . stringify the best way to get a deep copy in JS nowadays 😱

nin/dasBoot/NodeManager.js
```diff
@@ -5,22 +5,29 @@ class NodeManager {
...
     }
 
-    return new nodeConstructor(nodeInfo.id, nodeInfo.options);
+    return new nodeConstructor(deepClonedNodeInfo.id, deepClonedNodeInfo.options);
   }
 
   insertOrReplaceNode(node) {
     const nodeToReplace = this.nodes[node.id];
     if(nodeToReplace) {
+      for(let inputKey in nodeToReplace.inputs) {
```
aleksanb: Did this just never happen before?

aleksanb: `in` vs `of`? If inputs is an object, of is preferable, right?

nin/dasBoot/NodeManager.js
sigvef: It's not the worst, at least.

nin/dasBoot/NodeManager.js
sigvef: It did before-before, but then a recent change fixed some other bug and introduced this regression.

sigvef: We need the key explicitly, so we can't really use `of` here.

PR #470, opened 2017-12-15
Add interpolated BEAN number for animation use

lionleaf:
Adds new global variable INTERBEAN, which is a float that is
interpolated between BEAN numbers, in a linear way between frames.
This can be used to key animations without getting the lower frame rate
of BEAN.

Discussion to be had here: Can we instead change BEAN to have fractional values? The way I see it that would be the preferred option here, but I'm afraid of changing behaviour if people have already relied on it being an integer somewhere.

### DIFF
--- a/nin/dasBoot/BEATBEAN.js
+++ b/nin/dasBoot/BEATBEAN.js
@@ -26,7 +26,12 @@ const updateBeatBean = function(frame) {
       ((frame + 0.5) / framesPerBEAT) | 0) {
     window.BEAT = true;
   }
-  window.BEAN = (frame + 1.5) / framesPerBEAT | 0;
+  var bean = (frame + 1.5) / framesPerBEAT | 0;
+  window.BEAN = bean;
+
+  var f = (frame - FRAME_FOR_BEAN(bean))
+        / (FRAME_FOR_BEAN(bean+1) - FRAME_FOR_BEAN(bean));
+  window.INTERBEAN = lerp(bean, bean+1, f);
 };
 
 module.exports = {initBeatBean, updateBeatBean};


lionleaf:
INTERBEAN is also a bit long name for a variable you might repeat a ton of times while writing animations, anyone have a better suggestion?  FBEAN? (float bean)  BEAF?

sigvef (changes requested):
I imagine INTERBEAN might be piecewise linear from bean to bean, but have different slopes depending the relation between bpm and fps. Perhaps a better approach would be to construct INTERBEAN by just scaling frame to match? Of course, then INTERBEAN and BEAN wouldn't necessarily be equal when BEAT is true, which might not be optimal either 🤔

lionleaf:
Not sure I follow. This PR would make INTERBEAN piecewise linear with the slope depending on bpm and fps already, right?

sigvef:
There is no guarantee that there is a fixed number of frames in each BEAN. In this example, bpm is 125, subdivision is 4 and fps is 60.

<pre align="center"></pre>

Not a completely straight line:
<pre align="center"></pre>

So, if you were to use INTERBEAN e.g. like this:

```js
this.ball.position.x = Math.sin(INTERBEAN);
```

then the animation will speed up and slow down ever so slightly all the time.

lionleaf:
Yes, we agree about the current behaviour then. It is piecewise linear and certainly not just a straight line.

I went back to find the original use case that motivated this:

https://github.com/ninjadev/revision-invite-2018/blob/5b66355f75b18e894dcfcd2d257420f4dbee99d9/src/OrthoCubeNode.js#L29

Basically it's used for animations between states that complete on a beat. (In this case the rotation of the cube).

And since there could be different amounts of frames between beats you do want the rotation speed to vary.

You're right though, this could easily be used in scenarios where you want/expect a constant speed. 

An alternative that would also solve the motivating scenario would be a number that goes 0.0 -> 1.0 between each beat?

lionleaf:
Also happy to make the decision that this is more confusing than useful and close this PR.

sigvef:
Maybe we can just give it a name that implies the piece-wise-ness strongly enough and merge that.

PR #494, opened 2018-10-13
Adds package-lock to gitignore

peterhgombos:
Nin uses yarn, so the package-lock file from npm is not needed. This will stop it from being checked in at a later stage.


sigvef:
Maybe we should switch to npm since we use `npm link` anyway (and there is no yarn equivalent afaik)

PR #496, opened 2018-10-23
Use beforeUpdate instead of update by default in SceneSwitcherNode.

iver56:
This fixes a graph rendering delay issue.


iver56:
Perhaps we should also do this:
https://github.com/iver56/neon-fantasy/pull/17/files

stianjensen:
Good idea!

iver56:
🐹

PR #505, opened 2018-10-30
25.0.0

stianjensen:


stianjensen:
Turns out I've been logged out from npm on my laptop, and don't remember the npm login for the ninjadev account. So I haven't been able to publish this. Do you have the login @aleksanb ?

aleksanb:
I might. Can have a check later.

PR #508, opened 2018-12-02
Bump to version 27 in order to trigger new release

stianjensen:


stianjensen:
> Skipping a deployment with the npm provider because a custom condition was not met

Meh. Back to drawing board.

